PR #3, opened 2019-08-02
Different API Change.

fulmicoton:

PR #7, opened 2019-08-02
Tests

petr-tik:
added tests for AND BooleanQuery and ValueError in case of a malformed query

Moved document index setup into a classmethod.
classmethod will setup once at the start,
giving all test methods attributes to access.

I think it cuts on boilerplate in each test method


tests/tantivy_test.py
```diff
@@ -1,9 +1,15 @@
...
 import tantivy
 
+import pytest
+
 
 class TestClass(object):
-    def test_simple_search(self):
+
+    @classmethod
+    def setup_class(cls):
```
poljar: The usual way to achieve this is to setup a pytest fixture. That way you could also have multiple index fixtures for differing schemas for example. More info [here](https://docs.pytest.org/en/latest/fixture.html).

tests/tantivy_test.py
```diff
@@ -123,3 +123,54 @@ def test_query_errors(self):
         # no "bod" field
         with pytest.raises(ValueError):
             q_parser.parse_query("bod:title")
+
+
+class TestFromDiskClass(object):
+
+    @classmethod
+    def setup_class(cls):
```
poljar: Again, I think a fixture would be more idiomatic and more flexible.

petr-tik:
@poljar thanks for your review and recommending fixtures. 

I have tried my best in replacing classmethods with pytest fixture. 

Since @fulmicoton is changing the API with #3 we can either merge this now and enable Paul to test his API changes against some tests or wait for #3 to land and refactor/extend the tests.

fulmicoton:
@poljar I'm not sure if that was clear : you have write rights to this repository. I let you admnistrate the PR on this.

poljar:
I understand that i have write rights, It's still nice if someone else takes a look before I merge my own changes.

Anyways, this looks good now, gonna merge.

PR #11, opened 2019-08-29
Adding Field and Term to enable deleting docs from the index

bloodbare:


src/term.rs
```diff
@@ -0,0 +1,57 @@
...
+///
+/// This term can be used to access index and operate on that value
+/// for example index.delete_term(term) to delete all elements with that value
+#[pyclass]
+pub(crate) struct Term {
+    pub(crate) inner: schema::Term,
+}
+
+#[pymethods]
+impl Term {
```
fulmicoton: In the spirit of my changes, could we keep remove `Term` and `Field` and use a field_name in place?

Instead, we could have `.delete_term(field_name: &str, field_value: dynamicpythontype)` and do the resolution of the field name as well as the value to the internal schema defined type happen transparently when `.delete_term` is called.

bloodbare: I pushed an update applying your criteria, I'm using Document `extract_value`to avoid repeating code, maybe a utils module would be usefull to store this utility functions.

src/document.rs
```diff
@@ -109,7 +109,7 @@ where
         .push(Value::from(value));
 }
 
-fn extract_value(any: &PyAny) -> PyResult<Value> {
+pub fn extract_value(any: &PyAny) -> PyResult<Value> {
```
poljar: This can be a `pub(crate)`, no?

src/index.rs
```diff
@@ -94,6 +95,37 @@ impl IndexWriter {
     fn commit_opstamp(&self) -> u64 {
         self.inner_index_writer.commit_opstamp()
     }
+
+    /// Delete all documents containing a given term.
```
poljar: This needs documentation for the arguments and the return value. We also may want to figure out a better name for the method since we don't pass a term but a value.

src/index.rs
```diff
@@ -94,6 +95,37 @@ impl IndexWriter {
...
+    fn delete_term(
+        &mut self,
+        field_name: &str,
+        field_value: &PyAny,
+    ) -> PyResult<u64> {
+        if let Some(field) = self.schema.get_field(field_name) {
+            let value = extract_value(field_value).unwrap();
```
poljar: This unwrap needs to be gone. You can just propagate the error using the `?` operator. Don't forget to document that the method can raise exceptions.

bloodbare:
Moving to PR#12

PR #12, opened 2019-09-05
Adding delete_document function to index

bloodbare:


src/index.rs
```diff
@@ -94,6 +95,44 @@ impl IndexWriter {
...
+    ) -> PyResult<u64> {
+        if let Some(field) = self.schema.get_field(field_name) {
+            let value = extract_value(field_value)?;
+            let term = match value {
+                Value::Str(text) => Term::from_field_text(field, &text),
+                Value::U64(num) => Term::from_field_u64(field, num),
+                Value::I64(num) => Term::from_field_i64(field, num),
+                Value::F64(num) => Term::from_field_f64(field, num),
+                Value::Date(d) => Term::from_field_date(field, &d),
+                Value::Facet(facet) => Term::from_facet(field, &facet),
```
fulmicoton: No need to address :

I think this is ok for the moment. This approach will not work if we want to handle auto casting `str` to `Facet` as you suggested in the other PR though.

To do that we need to introduce something like a `fn convert_to_fieldvalue(entry: &FieldEntry, value: &Value) -> FieldValue` or something like that. The logic probably already exists somewhere since we parse Json with it.

src/index.rs
```diff
@@ -94,6 +95,44 @@ impl IndexWriter {
...
+                Value::Bytes(_) => {
+                    return Err(exceptions::ValueError::py_err(format!(
+                        "Field `{}` is bytes type not deletable.",
+                        field_name
+                    )))
+                }
+            };
+            Ok(self.inner_index_writer.delete_term(term.clone()))
+        } else {
+            return Err(exceptions::ValueError::py_err(format!(
```
fulmicoton: no need for `return` I think.

src/index.rs
```diff
@@ -94,6 +95,44 @@ impl IndexWriter {
...
+    ///     field_value (PyAny): Python object with the value we want to filter.
+    ///
+    /// If the field_name is not on the schema raises ValueError exception.
+    /// If the field_value is not supported raises Exception.
+    fn delete_documents(
+        &mut self,
+        field_name: &str,
+        field_value: &PyAny,
+    ) -> PyResult<u64> {
+        if let Some(field) = self.schema.get_field(field_name) {
```
fulmicoton: I tend to like `.ok_or_else` for that kind of stuff nowadays (I know my code is inconsistent on that point).
It removes one level and indentation and matches the way people would explain what happens.

```rust
let field = self.schema.get_field(field_name)
               .ok_or_else(|| exceptions::ValueError::py_err(format!(...)))?;
...

File: tests/tantivy_test.py [hunk 4]
```diff
@@ -103,6 +103,26 @@ def test_query_errors(self, ram_index):
...
+class TestUpdateClass(object):
+
+    def test_delete_update(self, ram_index):
+        query = ram_index.parse_query("Frankenstein", ["title"])
+        top_docs = tantivy.TopDocs(10)
+        result = ram_index.searcher().search(query, top_docs)
+        assert len(result) == 1
+
+        writer = ram_index.writer()
+        writer.delete_documents('title', "frankenstein")
```
fulmicoton: *nitpick* maybe use the same quotes for fields ands the value to avoid confusing users.

src/index.rs
fulmicoton: This is largely a matter of taste... so feel free to not do anything and makr the comment as resolved.

tests/tantivy_test.py
```diff
@@ -103,6 +103,26 @@ def test_query_errors(self, ram_index):
...
+        writer.commit();
+        ram_index.reload()
+
+        query = ram_index.parse_query("Frankenstein", ["title"])
```
fulmicoton: you can reuse the query above

tests/tantivy_test.py
```diff
@@ -103,6 +103,26 @@ def test_query_errors(self, ram_index):
...
+        top_docs = tantivy.TopDocs(10)
```
fulmicoton: you can reuse the top_docs above.

Issue #10, opened 2019-08-07
Support ordering search results by a field.


poljar:
Tantivy supports ordering the search results from a TopCollector by a field.
This used to be done with the `TopDocsByField` collector and a patch was written to partially utilize this https://github.com/matrix-org/tantivy/commit/85b8d0c7826092b11d81aa1f0b45e5512305f69c.

The latest Tantivy release dropped `TopDocsByfield` and uses the generic Collector trait instead.

This sadly makes it impossible to modify the mentioned patch since it uses the `Any` type and down-casting and down-casting trait objects is impossible:

     error: the `downcast_ref` method cannot be invoked on a trait object

There are tricks to work around of this but support for this would need to land in Tantivy.

bloodbare:
Would be great that `search` function allows a list of collectors. From the API point of view its better to map Tantivy collectors to Python objects and send them on a list or map each collector to a string so the API would be:

`searcher().search(query, 'top')`

`searcher().search(query, ['top', 'count'])`

`searcher().search(query, ['top', 'count', { 'person': ['john']}])`

or:

```
top_collector = tantivy.TopDocs(10)
count_collector = tantivy.Count()
facet_collector = tantivy.FacetDocs('/person/john', '/person/bob')
index.searcher().search(query, (top_collector, count_collector, facet_collector))
```

fulmicoton:
Yeah or even drop the idea of a collector here, and just add a bunch of options describing what we want.

e.g
- sort_by="bm25"
- count=True
-nhits=10
-facets=[]

We most likely won't handle custom collectors in python anyway, so we might as well have a specialized API that is more straightforward to understand.

poljar:
This has been put on hold since a wrongfully configured schema causes Tantivy to panic. Python users will most likely perceive crashes as bugs, until we can convert the panic into a Python exception this will unlikely be added.

fulmicoton:
@poljar Agreed. This is a bad behavior for rust too. Let's fix that.

poljar:
This is now possible, relevant commit 094f8974ea100fbb1eca7aad4078cd650410e51f.

PR #16, opened 2019-10-06
[WIP] Search API improvements.

poljar:
This patch-set simplifies the search API and adds the ability to order the results by a given field like it was suggested in #10.

The first patch removes the concept of a collector and uses an optional argument to limit the search result.

The second patch adds another optional argument to the method that adds the ability to to order the search result by an unsigned field.

While all of this is pretty straightforward the second patch has a problem. To be able to order a search result by a field the field needs to be an unsigned field that is marked as a fast field. The API already allows this when adding such a field, the problem arises if the developer fails to mark the field as a fast one. 

The search will crash the Python interpreter when it is run without a fast field. Tantivy raises a panic under such conditions and sadly the `Searcher` isn't unwind-safe so catching the panic and turning it into a `Result` and therefore turning it into a nice Python exception isn't possible.

Since crashes should be avoided under all circumstances on the Python side of things we a couple possibilities here:
- Don't let Python users create non-fast unsigned fields
- Modify Tantivy to return a Result type instead of calling panic.
- Don't expose the order by a fast field feature at all.

### DIFF
--- a/src/searcher.rs
+++ b/src/searcher.rs
@@ -2,11 +2,12 @@
 
 use crate::document::Document;
 use crate::query::Query;
-use crate::{to_pyerr, get_field};
+use crate::{to_pyerr};
+use pyo3::exceptions::ValueError;
 use pyo3::prelude::*;
 use pyo3::PyObjectProtocol;
 use tantivy as tv;
-use tantivy::collector::{MultiCollector, Count, TopDocs};
+use tantivy::collector::{Count, MultiCollector, TopDocs};
 
 /// Tantivy's Searcher class
 ///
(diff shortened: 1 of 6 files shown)


src/searcher.rs
```diff
@@ -29,20 +32,30 @@ impl Searcher {
     /// search results.
     ///
     /// Raises a ValueError if there was an error with the search.
+    #[args(limit = 10, sort_by = "SORT_BY")]
     fn search(
```
fulmicoton: We cannot have the matching document count with this setting though :-S

poljar: By matching count you mean using the `CountCollector`? If so, yeah I'm aware, adding that would probably mean switching to the `Multicollector`. I will probably do so at the next iteration of this.

bloodbare:
If its useful I've needed to implement a fast search function with MultiCollector for facets with a count collector and the top collector:

https://github.com/bloodbare/tantivy-py/blob/api-change/src/searcher.rs#L37

Its a really basic aproach and these days I don't have spare time to pulish and to a proper PR so if you want to use some of it feel free.

PR #18, opened 2019-12-17
Using tantivy 0.11

fulmicoton:


src/document.rs
```diff
@@ -50,6 +54,10 @@ fn value_to_string(value: &Value) -> String {
         Value::Bytes(bytes) => format!("{:?}", bytes),
         Value::Date(d) => format!("{:?}", d),
         Value::Facet(facet) => facet.to_string(),
+        Value::PreTokStr(pretok) => {
+            // TODO implement me
```
poljar: Can't we just use `pretok.text.clone()` here? I know that this isn't strictly required since we don't have the ability to add them but this seems to be easy to implement?

poljar (approved):
Test failing is the one with the old index format. There's somewhere in my tree a test that uses a temp dir instead so we can use that to test it with a non RAM directory.

Looks good.

PR #22, opened 2020-09-05
support for offset to do record pagination

sourcepirate:
Updates to searcher.rs to support extra arguments for python interface `search` function.

### DIFF
--- a/src/searcher.rs
+++ b/src/searcher.rs
@@ -104,7 +104,7 @@ impl Searcher {
         _py: Python,
         query: &Query,
         limit: usize,
-        offset: usize, 
+        offset: usize,
         count: bool,
         order_by_field: Option<&str>,
     ) -> PyResult<SearchResult> {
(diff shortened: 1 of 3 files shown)


src/searcher.rs
```diff
@@ -98,12 +98,13 @@ impl Searcher {
     /// Returns `SearchResult` object.
     ///
     /// Raises a ValueError if there was an error with the search.
-    #[args(limit = 10, count = true)]
+    #[args(limit = 10, offset = 0, count = true)]
```
poljar, line 101: The new argument needs to be documented.

src/searcher.rs
```diff
@@ -98,12 +98,13 @@ impl Searcher {
...
     fn search(
         &self,
         _py: Python,
         query: &Query,
         limit: usize,
+        offset: usize,
```
poljar, line 107: Changing the order of the fields here is a breaking change, any reason why we don't want to put it at least after count?

poljar (changes requested):
Generally looks good, needs some docs and if possible a test showing that this works as expected would be nice.

src/searcher.rs
sourcepirate: Got it!

sourcepirate:
updated code based on the comments. Let me know if it is good enough.

src/searcher.rs
sourcepirate: done!

sourcepirate:
@poljar any updates here ?

poljar:
> @poljar any updates here ?

A bit patience please, I'll look at this later today. Taking a glance at this, we probably should check in the test that we got the correct document when using the offset.

poljar:
I merged this but removed two commits that swapped the Tantivy version back and forth.

sourcepirate:
Awesome! thanks @poljar

PR #23, opened 2020-09-19
tantivy-py: Upgrade PyO3.

poljar:
This removes our nightly requirement but sadly it adds a requirement for things that are kept inside a Python class to be Send. Luckily for us almost everything in Tantivy is Send, except for the Query trait.

This patch works around this by keeping the parser and query string inside our python Query object. This sadly means that we are going to parse the query string twice.


fulmicoton:
I am ok to make Query send in tnativy and publish a 13.1. Would that work for you? Do you need another trait to be Send?

poljar:
> I am ok to make Query send in tnativy and publish a 13.1. Would that work for you? Do you need another trait to be Send?

Don't think so, nothing else needed any changes, well except for the renaming of `ValueError` to `PyValueError`. Here's the exact error:

```
#[pyclass]
^^^^^^^^^^ `(dyn tantivy::query::query::Query + 'static)` cannot be sent between threads safely
```

If marking the `Query` trait as `Send` is possible, doing so would be great. Otherwise we need this hack and parse a query twice.

Thanks.

fulmicoton:
@poljar 0.13.1 published. I added Sync+Send to both collector and queries.

poljar:
> @poljar 0.13.1 published. I added Sync+Send to both collector and queries.

Thanks, PR updated accordingly.

I'll prepare a release today or tomorrow.

fulmicoton:
LGTM!

PR #28, opened 2021-02-22
Upgrade pyo3 to 0.13 and build on stable Rust

messense:


fulmicoton:
Merged as a different PR as there was modification needed in the travis configuration.

PR #31, opened 2021-05-01
Release GIL on searcher acquisition.

moriyoshi:
I'm not really sure if this is appropriate, but the searchers queue is a blocking queue so it should be a good idea to give another thread a chance to run here.


poljar:
This certainly makes sense, the Rust side is thread safe so it should be ok. Has this been tested at all?

moriyoshi:
Yes, this patch got it to work in the expected manner. It used to end up in a mysterious pause in a multithreaded environment.

poljar:
Thanks, merged.

PR #34, opened 2021-10-17
Bump tantivy version 0.16.0

ccoder64:


fulmicoton:
Thank you @ccoder64 !

PR #36, opened 2021-11-16
Add SnippetGenerator for highlighting

jgreene:
Add SnippetGenerator for getting highlighted snippets from documents

Also updates to tantivy 0.16.1 and fixes the build by updating Cargo.toml.


src/searcher.rs
```diff
@@ -196,8 +196,8 @@ impl Searcher {
 #[pyclass]
 #[derive(Clone, Debug)]
 pub(crate) struct DocAddress {
-    pub(crate) segment_ord: tv::SegmentLocalId,
-    pub(crate) doc: tv::DocId,
+    pub(crate) segment_ord: tv::SegmentOrdinal,
+    pub(crate) doc_id: tv::DocId,
```
kapilt, line 200: any reason for changing the public interface here? that has stability implications for extant users.

jgreene: Since it is pub(crate) it should only impact the crate itself (I think).  I just updated the naming to match the tantivy changes.  It doesn't matter to me either way.

kapilt: ic, thanks for clarifying and verified.

Cargo.toml
```diff
@@ -22,4 +22,4 @@ features = ["extension-module"]
 
 [package.metadata.maturin]
 requires-python = ">=3.7"
-project-url =  { Source = "https://github.com/quickwit-inc/tantivy-py" }
+project-url =  ["https://github.com/quickwit-inc/tantivy-py"]
```
kapilt, line 25: afaics this change breaks building with maturin tooling

```
❯ maturin build
💥 maturin failed
  Caused by: Failed to parse Cargo.toml at Cargo.toml
  Caused by: invalid type: sequence, expected a map for key `package.metadata.maturin.project-url` at line 23 column 1
```

fulmicoton: @jgreene Can you confirm and fix?

src/snippet.rs
```diff
@@ -0,0 +1,69 @@
...
+        let text: String = doc
+            .iter_values_for_field(&self.field_name)
+            .flat_map(tv::schema::Value::text)
+            .collect::<Vec<&str>>()
+            .join(" ");
+
+        let result = self.inner.snippet(&text);
+        Snippet { inner: result }
+    }
+}
```
fulmicoton, line 69:
```suggestion
}

```

tests/tantivy_test.py
```diff
@@ -322,3 +322,27 @@ def test_document_with_facet(self):
...
+        for (score, doc_address) in result.hits:
+            doc = searcher.doc(doc_address)
+            snippet = snippet_generator.snippet_from_doc(doc)
+            highlights = snippet.highlighted()
+            assert len(highlights) == 1
+            first = highlights[0]
+            assert first.start == 20
+            assert first.end == 23
+            html_snippet = snippet.to_html()
+            assert html_snippet == 'The Old Man and the <b>Sea</b>'
```
fulmicoton, line 348:
```suggestion
            assert html_snippet == 'The Old Man and the <b>Sea</b>'

```

fulmicoton (approved):
LGTM but @kapilt made a valid comment about the Cargo.toml change.

cjrh:
Many thanks @jgreene, after eons your work has eventually been merged in #122 🙏🏼

PR #40, opened 2022-01-16
Update to pyo3 0.15

messense:


messense:
Wow, pipenv is so slow.

PR #41, opened 2022-01-17
Switch from pipenv to nox

messense:
pipenv is way too slow.

nox: https://nox.thea.codes/


ChillFish8:
Any particular reason for using nox over say Poetry?

messense:
Well, this project uses maturin so it does not really need another PEP 517 build backend and [Poetry doesn't support PEP 621](https://github.com/python-poetry/poetry/issues/3332).

ChillFish8:
> Well, this project uses Maturin so it does not really need another PEP 517 build backend and Poetry doesn't support PEP 621.
Fair enough

ChillFish8 (approved):
Looks fine to me, although If we ever add standard dependencies (although very unlikely) we'll want a lock file present (not sure if nox provides that for non-dev deps)

PR #45, opened 2022-04-15
Bump tantivy version 0.17.0

saukymo:


fulmicoton:
@ChillFish8  could you review?

src/document.rs
```diff
@@ -15,9 +15,37 @@ use tantivy as tv;
...
 
+fn value_to_object(val: &JsonValue, py: Python<'_>) -> PyObject {
+    match val {
+        JsonValue::Null => py.None(),
+        JsonValue::Bool(b) => b.to_object(py),
+        JsonValue::Number(n) => {
+            let oi64 = n.as_i64().map(|i| i.to_object(py));
+            let ou64 = n.as_u64().map(|i| i.to_object(py));
+            let of64 = n.as_f64().map(|i| i.to_object(py));
+            oi64.or(ou64).or(of64).expect("number too large")
```
ChillFish8, line 33: I feel like there is probably a cleaner way to do this, unfortunately, im unable to test this right now.

ChillFish8:
I'll try review this over the weekend, my computer's decided to call it quits right now so I'm unable to test this pr until then.

tests/tantivy_test.py
```diff
@@ -322,3 +330,66 @@ def test_document_with_facet(self):
...
+
+        query = index.parse_query("target:submit-button", ["attributes"])
+        result = index.searcher().search(query, 2)
+        assert len(result.hits) == 2
+
+        query = index.parse_query("target:submit", ["attributes"])
+        result = index.searcher().search(query, 2)
+        assert len(result.hits) == 2
+
+        # query = index.parse_query_for_attributes("cart.product_id:103")
```
saukymo, line 387: Comment out this due to issue: https://github.com/quickwit-oss/tantivy/issues/1348

src/document.rs
saukymo: I just copied code from this answer: https://stackoverflow.com/questions/70193869/exporting-hashmap-of-hashmap-to-python

ChillFish8: I think we probably want to try make this lazy if we can, bexause right now we're potentially calling python 3 or 4 times or so for a single value, which is quite high!

saukymo: How about now? Checked with is_i64, is_u64 & is_f64 and then turn to object py. In fact, I'm not sure if it's necessary to use `map(|i| i.to_object(py))` like before.

saukymo:
Format with Rustfmt and push again. thx.

ChillFish8 (approved):

src/document.rs
```diff
@@ -14,10 +14,37 @@ use chrono::{offset::TimeZone, Datelike, Timelike, Utc};
...
 
+fn value_to_object(val: &JsonValue, py: Python<'_>) -> PyObject {
+    match val {
+        JsonValue::Null => py.None(),
+        JsonValue::Bool(b) => b.to_object(py),
+        JsonValue::Number(n) => match n {
+            n if n.is_i64() => n.as_i64().to_object(py),
+            n if n.is_u64() => n.as_u64().to_object(py),
+            n if n.is_f64() => n.as_f64().to_object(py),
+            _ => panic!("number too large"),
```
ChillFish8, line 32: Sorry, I know this is slightly pedantic, but I'm pretty sure you can match on the num variant i.e `Number::I64`  which should remove the need for this panic. I think it'll be more idiomatic that way. :)

saukymo: The enum inside Number is a enum of 

``` Rust
enum N {
    PosInt(u64),
    /// Always less than zero.
    NegInt(i64),
    /// Always finite.
    Float(f64),
}

```
And it's a private field. So I don't figure out how to match them directly.

saukymo: Thank you! But I don't have the write access. Please help to merge it.

ChillFish8: 👍

PR #53, opened 2022-07-29
Export tantivy query classes

VsevolodZakharov:


Cargo.toml
```diff
@@ -11,15 +11,17 @@ name = "tantivy"
 crate-type = ["cdylib"]
 
 [build-dependencies]
-pyo3-build-config = "0.16.3"
+pyo3-build-config = "0.15.0"
```
kapilt, line 14: why re going backwards here on version?

cjrh:
These all appear to already have been added, so closing.

PR #56, opened 2022-08-05
Include Python 3.11 in nox tests and Github actions matrix

cjrh:
I tried running the nox tests against the latest Python 3.11 RC from the deadsnakes ppa, and it worked.  Hopefully it also works under Github actions.  This is in preparation of the release of 3.11 which happens in October.


cjrh:
I see that nox is using the python version as defined in the matrix.  I'll have to figure out how to disentangle that.

image

.github/workflows/ci.yml
```diff
@@ -25,13 +25,20 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         python-version: [3.9]
+        majorminor: ["3.9"]
```
cjrh, line 28: A bit tedious but it should work?

This also opens the door to easily testing other Python distributions like PyPy, for which we'll have the same issue because the GA setup-python name won't match the nox major.minor python version session spec..

cjrh:
Since Python 3.11 has been released, the change is much simpler and just adds 3.11 into the matrix.

PR #57, opened 2022-08-07
Add python release build

wallies:


wallies:
So what this does is adds SHAs as the github actions which is one of the steps the open security foundation. I used https://app.stepsecurity.io/ for this as this is the recommended approach by openssf. 

The other thing this does is add a publish step on workflow dispatch, so manual run. We can change it to be based on tags if you wish

fulmicoton:
Thank you! let me see if I can find someone to review this

kapilt:
Re action locking, I’d rather see use of open source for security purposes.. https://github.com/sethvargo/ratchet is one such alternative. Use of such tool should be in its own pr, no commingled. It also avoids polluting every single job with extraneous steps. I’ll also note while it is a good practice, it is very rare in the wild to see action locking.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,178 @@
...
+          args: --release -o dist --universal2
+
+      - name: Upload wheels
+        uses: actions/upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8  # 3.1.0
+        with:
+          name: wheels
+          path: dist
+
+      - name: Upload release binaries
+        uses: alexellis/upload-assets@eaab1472d5c2f0df3115cac81136f03e7e61f771
```
kapilt, line 144: Using the pypa organization’s upload action is preferred

kapilt:
It feels odd doing automation on a single old version of python on a single set of architecture, ideally this would be a matrix build with arch support for at least x86_64 and aarch64

kapilt:
> I used https://app.stepsecurity.io/ for this as this is the recommended approach by openssf.

could you clarify where you see/found this recommendation?, the only reference I can find to it, is a pr announce from step security itself.

wallies:
> > I used https://app.stepsecurity.io/ for this as this is the recommended approach by openssf.
> 
> could you clarify where you see/found this recommendation?, the only reference I can find to it, is a pr announce from step security itself.

https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies

Sidhant29:
@wallies is this PR good to go ?

wallies:
@Sidhant29 no needs to pull across changes from Kapiche fork

kapilt:
Lgtm, the tooling on step security is both helpful and easily replaced.

cjrh (approved):

PR #65, opened 2023-01-17
Add Float Support

Sidhant29:
Hi,
I have added a function `add_float_field()` that allows adding a float field to the schema. I have generated wheels from this branch and tested searching and adding to the field. I have also added some tests to confirm the desired functionality.


src/document.rs
```diff
@@ -292,6 +292,15 @@ impl Document {
         add_value(self, field_name, value);
     }
 
+    /// Add a float value to the document.
+    ///
+    /// Args:
+    ///     field_name (str): The field name for which we are adding the integer.
```
cjrh, line 298:
```suggestion
    ///     field_name (str): The field name for which we are adding the value.
```

cjrh (changes requested):
Looks good, but fix the incorrect text "integer"

cjrh (approved):

PR #67, opened 2023-02-05
Tantivy 0.19.2

pdavies011010:
Upgrade all Cargo dependencies, including upgrade to Tantivy 0.19.2
Build in customer tokenizers so other supported languages can be used


pratyushmittal:
Tantivy 0.19.1 version fixes so many bugs vs the current 0.17. The incremental indexes were increasing index size exponentially till v0.17.

Thanks a lot @pdavies011010 for adding support for the newer version.

pdavies011010:
Thanks @pratyushmittal ! I'm admittedly not a Rust expert so please let me know if there's anything I should do to help get this merged in. One thing that's hopefully okay is I switched the makefile to use maturin for the build, that's the way that has been working best for me to create builds wheels for multiple python3 versions, if not okay I can back that change out.

src/schemabuilder.rs
```diff
@@ -203,7 +204,27 @@ impl SchemaBuilder {
...
+                    "single" => Some(schema::Cardinality::SingleValue),
+                    "multi" => Some(schema::Cardinality::MultiValues),
+                    _ => return Err(exceptions::PyValueError::new_err(
+                        "Invalid index option, valid choices are: 'multivalue' and 'singlevalue'"
+                    )),
+                }
+            }
+            None => None,
+        };
+        opts = opts.set_fast(fast.unwrap());
```
Sidhant29, line 227: Isn't this just `build_integer_option` expanded ?

Sidhant29, line 227:
```suggestion
        let cardinality = match fast {
            Some(f) => {
                let f = f.to_lowercase();
                match f.as_ref() {
                    "single" => Some(schema::Cardinality::SingleValue),
                    "multi" => Some(schema::Cardinality::MultiValues),
                    _ => return Err(exceptions::PyValueError::new_err(
                        "Invalid index option, valid choices are: 'multivalue' and 'singlevalue'"
                    )),
                }
            }
            None => None,
        };
        if fast.is_some(){
            opts = opts.set_fast(fast.unwrap());
        }
```

I tried these changes out with some sample projects and this seems to fail in case fast is 'None', as 
`pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value`.

Apart from this looks good!

Sidhant29: Also you can format rust code using
`rustfmt --check src/*.rs`.

pdavies011010: @Sidhant29  Good finds, I updated to put in the fix you suggested, and applied rustfmt.

PSeitz:
tantivy `0.19.2` has been released

pdavies011010:
@PSeitz  Updated to Tantivy 0.19.2

PSeitz:
@ChillFish8 Can you have a look?

Cargo.toml
```diff
@@ -11,15 +11,15 @@ name = "tantivy"
...
-chrono = "0.4.19"
-tantivy = "0.17"
-itertools = "0.10.3"
-futures = "0.3.21"
-serde_json = "1.0.64"
+chrono = { version = "0.4.23" }
+tantivy = { version = "0.19.2" }
+itertools = { version = "0.10.5" }
+futures = { version = "0.3.26" }
+serde_json = { version = "1.0.91" }
```
ChillFish8, line 21: I dont think this change is worth it (using version = xxx) we don't use any of the features so we can just use the standard `dep = "x.x.x"` format.

rustfmt.toml
```diff
@@ -1 +1,2 @@
 max_width = 80
+edition = "2018"
```
ChillFish8, line 2: I don't think this is needed. It'll read the cargo file.

src/document.rs
```diff
@@ -58,13 +58,13 @@ fn value_to_py(py: Python, value: &Value) -> PyResult<PyObject> {
...
-            d.minute() as u8,
-            d.second() as u8,
-            d.timestamp_subsec_micros(),
+            d.into_utc().year(),
+            d.into_utc().month() as u8,
+            d.into_utc().day() as u8,
+            d.into_utc().hour() as u8,
+            d.into_utc().minute() as u8,
+            d.into_utc().second() as u8,
+            d.into_utc().microsecond() as u32,
```
ChillFish8, line 67: It is probably easier here to have once `into_utc()` call and then work on that result for all the values rather than calling it every time.

src/document.rs
```diff
@@ -95,6 +97,8 @@ fn value_to_string(value: &Value) -> String {
         Value::JsonObject(json_object) => {
             serde_json::to_string(&json_object).unwrap()
         }
+        Value::Bool(b) => format!("{}", b),
+        Value::IpAddr(i) => format!("{}", *i),
```
ChillFish8, line 101: Nitpick, clippy now recommends using the new formatting method `format!("{b}")` for example.

src/document.rs
```diff
@@ -387,9 +395,9 @@ impl Document {
     }
 
     fn __getitem__(&self, field_name: &str) -> PyResult<Vec<PyObject>> {
-        let gil = Python::acquire_gil();
-        let py = gil.python();
-        self.get_all(py, field_name)
+        return Python::with_gil(|py| -> PyResult<Vec<PyObject>> {
+            self.get_all(py, field_name)
+        });
```
ChillFish8, line 400: needless return
```suggestion
        Python::with_gil(|py| -> PyResult<Vec<PyObject>> {
            self.get_all(py, field_name)
        })
```

src/index.rs
```diff
@@ -355,3 +368,34 @@ impl Index {
...
+        analyzers.insert("hu_stem", Language::Hungarian);
+        analyzers.insert("it_stem", Language::Italian);
+        analyzers.insert("no_stem", Language::Norwegian);
+        analyzers.insert("pt_stem", Language::Portuguese);
+        analyzers.insert("ro_stem", Language::Romanian);
+        analyzers.insert("ru_stem", Language::Russian);
+        analyzers.insert("es_stem", Language::Spanish);
+        analyzers.insert("sv_stem", Language::Swedish);
+        analyzers.insert("ta_stem", Language::Tamil);
+        analyzers.insert("tr_stem", Language::Turkish);
```
ChillFish8, line 391: This doesn't need to be a hashmap here, an array would be fine.

src/lib.rs
```diff
@@ -1,5 +1,5 @@
+use ::tantivy as tv;
```
ChillFish8, line 1: I believe this is a needless change, feel free to correct me though.

ChillFish8 (changes requested):
Looks pretty good overall, just a few small things to sort.

rustfmt.toml
pdavies011010: So it will be read from Cargo.toml if you use `cargo fmt`, but not if you run `rustfmt` directly. If we want to standardize around using cargo fmt I can update the Makefile to use it instead of rustfmt, but otherwise I think we want to leave this.

src/lib.rs
pdavies011010: Without that change I get this during the build: 
```
error[E0659]: `tantivy` is ambiguous=> ] 177/178: tantivy
  --> src/lib.rs:2:5
   |
2  | use tantivy as tv;
   |     ^^^^^^^ ambiguous name
   |
   = note: ambiguous because of multiple potential import sources
   = note: `tantivy` could refer to a crate passed with `--extern`
   = help: use `::tantivy` to refer to this crate unambiguously
note: `tantivy` could also refer to the module defined here
  --> src/lib.rs:69:1
   |
69 | #[pymodule]
   | ^^^^^^^^^^^
   = help: use `crate::tantivy` to refer to this module unambiguously
   = note: this error originates in the attribute macro `pymodule` (in Nightly builds, run with -Z macro-backtrace for more info)
```

rustfmt.toml
ChillFish8: I think standardising everything to use cargo would be a good idea 😃

src/lib.rs
ChillFish8: Ah, I see, it's getting confused with the function `tantivy` and the module, weird that PyO3 now errors on this when it didn't use to.

This is fine to keep then 👍

rustfmt.toml
pdavies011010: K, updated the Makefile and removed this line

Cargo.toml
pdavies011010: Done

src/document.rs
pdavies011010: Done

src/document.rs
pdavies011010: Linted with clippy

src/document.rs
pdavies011010: Done

src/index.rs
pdavies011010: Done

pdavies011010:
@ChillFish8 Cleaned things up per your suggestions, thanks!

ChillFish8 (approved):

ChillFish8:
Looks good 👍 Just need to run fmt to fix the CI and then this can be merged

pdavies011010:
Whoops, the CI config was also set to use rustfmt instead of `cargo fmt`, which is why the Lint step failed. I updated it (hopefully successfully) to use cargo fmt instead.

PR #82, opened 2023-07-02
feat: tantivy_0.20.1_upgrade

Sidhant29:
This Pr reflects the API changes observed from tantivy 0.19.2 to 0.20.1.

- The numeric_options interface has changes from requiring a str param of 'single', 'multivalued', to just a bool param. Fast fields hence support only 'single' mode.
- The `order_by_u64` method now accepts a 'str' field name, instead of the entire field object.
- The `schema.get_field` method returns a Result instead of an option.

### DIFF
--- a/src/searcher.rs
+++ b/src/searcher.rs
@@ -113,8 +113,6 @@ impl Searcher {
 
         let (mut multifruit, hits) = {
             if let Some(order_by) = order_by_field {
-                // let schema = &self.inner.index().schema();
-                // let field = get_field(schema, order_by)?;
                 let collector = TopDocs::with_limit(limit)
                     .and_offset(offset)
                     .order_by_u64_field(order_by);
--- a/src/searcher.rs
+++ b/src/searcher.rs
@@ -1,6 +1,6 @@
 #![allow(clippy::new_ret_no_self)]
 
-use crate::{document::Document, get_field, query::Query, to_pyerr};
+use crate::{document::Document, query::Query, to_pyerr};
 use pyo3::{exceptions::PyValueError, prelude::*};
 use tantivy as tv;
 use tantivy::collector::{Count, MultiCollector, TopDocs};
@@ -113,10 +113,11 @@ impl Searcher {
 
         let (mut multifruit, hits) = {
             if let Some(order_by) = order_by_field {
-                let field = get_field(&self.inner.index().schema(), order_by)?;
+                // let schema = &self.inner.index().schema();
+                // let field = get_field(schema, order_by)?;
                 let collector = TopDocs::with_limit(limit)
                     .and_offset(offset)
-                    .order_by_u64_field(field);
+                    .order_by_u64_field(order_by);
                 let top_docs_handle = multicollector.add_collector(collector);
                 let ret = self.inner.search(query.get(), &multicollector);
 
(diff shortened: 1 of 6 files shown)


src/searcher.rs
```diff
@@ -113,10 +113,11 @@ impl Searcher {
 
         let (mut multifruit, hits) = {
             if let Some(order_by) = order_by_field {
-                let field = get_field(&self.inner.index().schema(), order_by)?;
+                // let schema = &self.inner.index().schema();
```
adamreichold, line 116: Left-over comments?

tests/tantivy_test.py
```diff
@@ -418,8 +418,8 @@ def test_create_readers(self):
 
 class TestSearcher(object):
     def test_searcher_repr(self, ram_index, ram_index_numeric_fields):
-        assert repr(ram_index.searcher()) == "Searcher(num_docs=3, num_segments=1)"
-        assert repr(ram_index_numeric_fields.searcher()) == "Searcher(num_docs=2, num_segments=1)"
+        assert repr(ram_index.searcher()) == "Searcher(num_docs=3, num_segments=3)"
```
adamreichold, line 421: Why did this change and is the reason stable?

cjrh:
I had a look too, same question as @adamreichold why did the test change?

tests/tantivy_test.py
cjrh: I did a bisection (not easy, because of the ff changes).  The tantivy commit that introduces the change in this tantivy-py test above is this: https://github.com/quickwit-oss/tantivy/pull/1828/commits/bd5eea98527e74d7f57dd39d62e4ab25ce8a5e5b

Unfortunately it's the big columnar merge, 180 changed files :(

cjrh: We (tantivy-py) may just have to trust that tantivy is doing here with segments is correct. I don't think it's any kind of regression or change on our side.

PSeitz: > Number of threads would have been my first idea as well and are also why I asked about stability. To me it looks like for the `RamDirectory`, we always have `num_docs == num_segments`? Maybe @PSeitz has an idea why that might be the case?

Maybe the memory limit is too low, so it commits after every doc

cjrh: In this unit test only 3 tiny documents are added. Is it still possible a memory limit is involved? Is it configurable or do you have a link (to source code presumably) where I can read up more about it?

adamreichold: I think that the memory limit involved would be the one passed on [`Index::writer`](https://docs.rs/tantivy/latest/tantivy/struct.Index.html#method.writer) and stored in [`IndexWriter::memory_arena_in_bytes_per_thread`](https://docs.rs/tantivy/latest/src/tantivy/indexer/index_writer.rs.html#60).

adamreichold: So from

https://github.com/quickwit-oss/tantivy-py/blob/468166f334b27f11a4a158c7565454bf1b26986b/src/index.rs#L220

I would say the tests use `3_000_000`.

adamreichold: I had similar problem in the example in https://github.com/quickwit-oss/tantivy/pull/2130 and bumping the memory to `10_000_000` resolved things there.

cjrh: That was it!  Thx @adamreichold & @PSeitz.

src/searcher.rs
```diff
@@ -113,10 +113,11 @@ impl Searcher {
...
+                // let field = get_field(schema, order_by)?;
```
wallies, line 117:
```suggestion
```

cjrh (approved):

wallies (approved):

Issue #47, opened 2022-04-25
Document.from_dict doesn't has type info.


saukymo:
Test with below code:

```python
class TestUnsignedField:
    def test_query_from_unsigned_field(self):
        schema = SchemaBuilder().add_unsigned_field("order", indexed=True).build()

        index = Index(schema)

        writer = index.writer()

        doc = Document.from_dict({"order": 1})

        writer.add_document(doc)
        writer.commit()
        index.reload()

        query = index.parse_query("order:1", ["order"])
        result = index.searcher().search(query, 1)
        assert len(result.hits) == 1
```
The extract_value in document.rs will try to interpret it as i64 at first while it should be interpreted as u64. We may need to pass schema into from_dict.

I also met this issue when I try to support json field where it interpreted the field as String at first.

GodTamIt:
This can be closed, as it was fixed in #88 .

PR #88, opened 2023-07-08
Add schema validation to PyDict -> Document

GodTamIt:
This change addresses the issue mentioned here: https://github.com/quickwit-oss/tantivy-py/issues/47

`Document.extend()` and `Document.from_dict()` support an optional `schema` field that, when provided, validates the provided dictionary against the schema.

This additionally fixes the issue where all numeric values are first parsed as `I64` and upon failure, parsed as `F64`. This can cause problems for fields that are unsigned (`U64`) but are parsed as `I64` instead, or fields that are float (`F64`) but is parsed as `I64`. Now, when a `schema` is provided, values will be parsed according to the schema's field specification.

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -128,7 +128,25 @@ fn value_to_string(value: &Value) -> String {
 ///
 /// Example:
 ///     >>> doc = tantivy.Document(title="The Old Man and the Sea", body="...")
-
+///
+/// For numeric fields, the [`Document`] constructor does not have any
+/// information about the type and will try to guess the type.
+/// Therefore, it is recommended to use the [`Document::from_dict()`],
+/// [`Document::extract()`], or `Document::add_*()` functions to provide
+/// explicit type information.
+///
+/// Example:
+///     >>> schema = (
+///             SchemaBuilder()
+///                 .add_unsigned_field("unsigned")
+///                 .add_integer_field("signed")
+///                 .add_float_field("float")
+///                 .build()
+///         )
+///     >>> doc = tantivy.Document.from_dict(
+///             {"unsigned": 1000, "signed": -5, "float": 0.4},
+///             schema,
+///         )
 #[pyclass]
 #[derive(Default)]
 pub(crate) struct Document {
@@ -330,6 +348,11 @@ impl Document {
 
 #[pymethods]
 impl Document {
+    /// Creates a new document with optional fields from `**kwargs`.
+    ///
+    /// Note that the types of numeric fields are unknown here. To
+    /// provide explicit type information, use the [`from_dict()`],
+    /// [`extend()`], or `add_<type>()` functions.
     #[new]
     #[pyo3(signature = (**kwargs))]
     fn new(kwargs: Option<&PyDict>) -> PyResult<Self> {
--- a/src/document.rs
+++ b/src/document.rs
@@ -9,7 +9,7 @@ use pyo3::{
     },
 };
 
-use chrono::{offset::TimeZone, DateTime, Utc};
+use chrono::{offset::TimeZone, NaiveDateTime, Utc};
 
 use tantivy as tv;
 
@@ -175,7 +175,7 @@ pub(crate) fn extract_value(any: &PyAny) -> PyResult<Value> {
     if let Ok(num) = any.extract::<f64>() {
         return Ok(Value::F64(num));
     }
-    if let Ok(datetime) = any.extract::<DateTime<Utc>>() {
+    if let Ok(datetime) = any.extract::<NaiveDateTime>() {
         return Ok(Value::Date(tv::DateTime::from_timestamp_secs(
             datetime.timestamp(),
         )));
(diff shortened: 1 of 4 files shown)


src/document.rs
```diff
@@ -200,6 +200,71 @@ pub(crate) fn extract_value(any: &PyAny) -> PyResult<Value> {
...
+        ),
+        tv::schema::Type::I64 => Value::I64(
+            any.extract::<i64>()
+                .map_err(to_pyerr_for_type("I64", field_name, any))?,
+        ),
+        tv::schema::Type::F64 => Value::F64(
+            any.extract::<f64>()
+                .map_err(to_pyerr_for_type("F64", field_name, any))?,
+        ),
+        tv::schema::Type::Date => {
```
adamreichold, line 239: Did you try the upstream integration of `chrono`, available by enabling PyO3's [`chrono` feature](https://pyo3.rs/v0.19.1/features#chrono).

src/document.rs
```diff
@@ -200,6 +200,71 @@ pub(crate) fn extract_value(any: &PyAny) -> PyResult<Value> {
...
+                .inner,
+        ),
+        _ => return Err(to_pyerr(format!("Value unsupported {:?}", any))),
+    };
+
+    Ok(value)
+}
+
 fn extract_value_single_or_list(any: &PyAny) -> PyResult<Vec<Value>> {
     if let Ok(values) = any.downcast::<PyList>() {
```
adamreichold, line 269: Not related to this PR, but I think this would benefit from using `.extract::<Vec<Value>>()` or at least `.extract::<Vec<&PyAny>>()` instead of `.downcast::<PyList>()` to handle any sequence and not just lists.

src/document.rs
```diff
@@ -208,51 +273,135 @@ fn extract_value_single_or_list(any: &PyAny) -> PyResult<Vec<Value>> {
...
+    any: &PyAny,
+    field_type: &tv::schema::FieldType,
+    field_name: &str,
+) -> PyResult<Vec<Value>> {
+    // Check if a numeric fast field supports multivalues.
+    let allow_multivalues = match field_type {
+        tv::schema::FieldType::U64(opts)
+        | tv::schema::FieldType::I64(opts)
+        | tv::schema::FieldType::F64(opts) => opts
+            .get_fastfield_cardinality()
```
adamreichold, line 286: I think this will not work after upgrading this crate to Tantivy 0.20? Probably best to consider to #82.

GodTamIt: Ack. If that change is landing soon, I can wait to rebase changes on top of that.

src/document.rs
GodTamIt: Would this needlessly create a temporary `Vec`? I'm transforming/mapping the elements inside, so maybe it's not worth collecting until the very end?

src/document.rs
GodTamIt: I've now rebased this on top of #82 .

src/document.rs
adamreichold: > Would this needlessly create a temporary Vec? I'm transforming/mapping the elements inside, so maybe it's not worth collecting until the very end?

Yes, that would be the cost for supporting arbitrary sequences (like 1d NumPy arrays) instead of just lists. It is a trade-off and since it isn't really part of this PR it is probably best left to do elsewhere if done at all.

(Alternatively, this could use `any.iter()` to just iterate of the sequence, but `.iter()` also works for e.g. strings which would need to be checked before doing this.)

wallies:
@GodTamIt do you want to rebase again now that the 0.20.1 PR has been merged

src/document.rs
adamreichold: @GodTamIt I have noticed the review request and have this on my to-do list, but I wanted to wait until the bump to Tantivy 0.20 had landed, which is the case now. So I'll do this as soon as I find the time.

src/document.rs
```diff
@@ -208,51 +251,119 @@ fn extract_value_single_or_list(any: &PyAny) -> PyResult<Vec<Value>> {
...
-            }
-        }
-        self.field_values.extend(field_values.into_iter());
-        Ok(())
+    fn extend(
+        &mut self,
+        py_dict: &PyDict,
+        schema: Option<&Schema>,
+    ) -> PyResult<()> {
+        Document::extract_py_values_from_dict(
```
adamreichold, line 348: I am not sure if the extra indirection of `extract_py_values_from_dict` is really useful here.

adamreichold: Ah sorry, it is called twice. Nevermind.

src/schemabuilder.rs
```diff
@@ -233,21 +233,8 @@ impl SchemaBuilder {
         if indexed {
             opts = opts.set_indexed();
         }
-        let fast = match fast {
```
adamreichold, line 236: These changes look like they are from the Tantivy bump?

GodTamIt: Rebased.

adamreichold (approved):

GodTamIt:
Thanks! This should be squash-merged

PR #94, opened 2023-07-21
doc: add MkDocs documentation

cjrh:
# Overview

This PR adds a `/docs` directory with content generated using [MkDocs](https://www.mkdocs.org/) which is becoming something of a standard in the py community.  MkDocs is a pleasure to work with, and I hope encourages contributors to improve the documentation.

The goal is to deploy this documentation on https://readthedocs.org/.

I've added a file structure and tree based on the [Diátaxis](https://diataxis.fr/). I am by no means an expert in this framework, but the basic idea of separating different content based on the needs of the reader makes a lot of sense and seems tractable for us to maintain.

It is obviously quite sparse at the moment, but one has to start somewhere.  I've copied some of the information in the README into these docs, into where I thought were the appropriate sections.  If we're happy with these docs, and we get them deployed to readthedocs, then we can remove some of that duplicated information from the README and just link to the new docs.

# Screenshots

Here are some examples of what it looks like:

The home page that will show on first load:

image

Some of the tutorial content from our existing README:

image


wallies:
@cjrh amazing docs. What's the flow to deploy these docs?

cjrh:
> What's the flow to deploy these docs?

The flow is "I ask @wallies to set up something" and then magically some steps show up in CI to deploy to readthedocs 😁

cjrh:
@wallies @Sidhant29 Can we move forward on this now, and then set up deploy to readthedocs as a separate work?

wallies (approved):

PR #97, opened 2023-07-21
Support pickling of some objects

### LINKED ISSUE #96: Support pickling of certain types
I've found the need for this feature at times for things like `Document`.

Maybe using something like `bincode` and doing:

```python
pub fn __setstate__(&mut self, state: &PyBytes) -> PyResult<()> {
    *self = bincode::deserialize(state.as_bytes()).map_err(...);
    Ok(())
}

pub fn __getstate__<'py>(&self, py: Python<'py>) -> PyResult<&'py PyBytes> {
    Ok(PyBytes::new(py, bincode::serialize(&self).map_err(...))
}
```

Thoughts?

GodTamIt:
Fixes #96, supporting serialization via `pythonize`.

The (personal) use-case here is to be able to send types across process boundaries (i.e. large batches of `Document`s across to a process pool for some kind of processing).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,8 @@ chrono = "0.4.23"
 tantivy = "0.20.1"
 itertools = "0.10.5"
 futures = "0.3.26"
+pythonize = "0.19.0"
+serde = "~1.0"
 serde_json = "1.0.91"
 
 [dependencies.pyo3]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ tantivy = "0.20.1"
 itertools = "0.10.5"
 futures = "0.3.26"
 pythonize = "0.19.0"
-serde = "~1.0.181"
+serde = "1.0"
 serde_json = "1.0.91"
 
 [dependencies.pyo3]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ tantivy = "0.20.1"
 itertools = "0.10.5"
 futures = "0.3.26"
 pythonize = "0.19.0"
-serde = "~1.0"
+serde = "~1.0.181"
 serde_json = "1.0.91"
 
 [dependencies.pyo3]
--- a/src/document.rs
+++ b/src/document.rs
@@ -16,9 +16,7 @@ use tantivy::{self as tv, schema::Value};
 
 use crate::{facet::Facet, schema::Schema, to_pyerr};
 use serde::{
-    de::{MapAccess, Visitor},
-    ser::SerializeMap,
-    Deserialize, Deserializer, Serialize, Serializer,
+    ser::SerializeMap, Deserialize, Deserializer, Serialize, Serializer,
 };
 use serde_json::Value as JsonValue;
 use std::{
@@ -243,8 +241,12 @@ where
     i64::deserialize(deserializer).map(tv::DateTime::from_timestamp_nanos)
 }
 
-/// An equivalent type to [`tantivy::schema::Value`] but uses tagging in its serialization to
-/// differentiate between different integer types.
+/// An equivalent type to [`tantivy::schema::Value`], but unlike the tantivy crate's serialization
+/// implementation, it uses tagging in its serialization and deserialization to differentiate
+/// between different integer types.
+///
+/// [`BorrowedSerdeValue`] is often used for the serialization path, as owning the data is not
+/// necessary for serialization.
 #[derive(Deserialize, Serialize)]
 enum SerdeValue {
     /// The str type is used for any text information.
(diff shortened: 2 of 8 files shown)


cjrh:
Hi @GodTamIt 

Thank you for this! A few general comments:

The maintainers here have very little bandwidth, so larger PRs can take a lot of time to get reviewed. Don't get discouraged if it takes a while, this is normal. Generally speaking, smaller PRs travel exponentially faster than large PRs, in many repos not only this one.

The PR includes extra unnecessary changes like renaming `Value` to `TvValue`. I prefer large renames and similar refactors like this to be in separate PRs. (and if this particular rename was in its own PR, I would reject the PR as unnecessary).

I feel like the motivation for supporting pickle feels not yet strong enough to support a change of this size, and the taking on of new dependencies (in my opinion). A use-case would help to understand the motivation better.  Is it to be able to move Fruit between processes/machines? Or a distributed system or similar? If we had our documentation already in place I'd ask you to add docs about the pickle feature which would help to explain more, but the docs aren't quite ready yet (#94).

This PR adds new dependencies. This will add to the time for review because I will go look at each of the dependencies to look for reasons to reject them. I have a high bar for taking on new dependencies. They are easy to add and very hard to remove. For instance, flexbuffers is a heavy C++ dependency we take on, just to support pickle. I would like us to try harder to make bincode work, or pythonize as suggested in #96? Or perhaps [serde_pickle](https://docs.rs/serde-pickle/latest/serde_pickle/), etc.

I see you added the time crate, but seeing as we already have chrono, did you look at [chrono::serde](https://docs.rs/chrono/latest/chrono/serde/)?

GodTamIt:
> Hi @GodTamIt
> 
> Thank you for this! A few general comments:
> 
> The maintainers here have very little bandwidth, so larger PRs can take a lot of time to get reviewed. Don't get discouraged if it takes a while, this is normal. Generally speaking, smaller PRs travel exponentially faster than large PRs, in many repos not only this one.
>
> The PR includes extra unnecessary changes like renaming `Value` to `TvValue`. I prefer large renames and similar refactors like this to be in separate PRs. (and if this particular rename was in its own PR, I would reject the PR as unnecessary).

Ack. I appreciate the feedback here, and I've removed these changes. In the future, I will try to limit the change to only the necessary changes.

> I feel like the motivation for supporting pickle feels not yet strong enough to support a change of this size, and the taking on of new dependencies (in my opinion). A use-case would help to understand the motivation better. Is it to be able to move Fruit between processes/machines? Or a distributed system or similar? If we had our documentation already in place I'd ask you to add docs about the pickle feature which would help to explain more, but the docs aren't quite ready yet (#94).

The primary motivation here is to be able to send these types (`Document`, `SearchResult`, etc.) across a process boundary. For example, `concurrent.futures.ProcessPoolExecutor` to operate on huge batches of `Document`s.

> This PR adds new dependencies. This will add to the time for review because I will go look at each of the dependencies to look for reasons to reject them. I have a high bar for taking on new dependencies. They are easy to add and very hard to remove. For instance, flexbuffers is a heavy C++ dependency we take on, just to support pickle. I would like us to try harder to make bincode work, or pythonize as suggested in #96? Or perhaps [serde_pickle](https://docs.rs/serde-pickle/latest/serde_pickle/), etc.

Ah, I didn't see the `pythonize`suggestion until after posting this PR. I've elected for that, as it seems to be far lighter as a dependency and something that works.

> I see you added the time crate, but seeing as we already have chrono, did you look at [chrono::serde](https://docs.rs/chrono/latest/chrono/serde/)?

`tantivy` opts for the `time` crate internally instead. However, since `tantivy` likes to store it as a nanosecond timestamp, I've changed the code to just use an `i64`.

Cargo.toml
```diff
@@ -18,8 +18,10 @@ chrono = "0.4.23"
...
 itertools = "0.10.5"
 futures = "0.3.26"
+pythonize = "0.19.0"
+serde = "~1.0"
 serde_json = "1.0.91"
 
 [dependencies.pyo3]
-version = "0.18.0"
-features = ["chrono", "extension-module"]
+version = "0.19.1"
```
GodTamIt, line 26: This change needs to wait for #98 to land.

GodTamIt:
This change was stacked on top of #99 (for equality support) and will need #98 to land before it can go in.

Cargo.toml
cjrh: It's in 👍🏼

Cargo.toml
```diff
@@ -18,6 +18,8 @@ chrono = "0.4.23"
 tantivy = "0.20.1"
 itertools = "0.10.5"
 futures = "0.3.26"
+pythonize = "0.19.0"
+serde = "~1.0"
```
adamreichold, line 22: Why a non-default version specification for the requirement on `serde`?

src/document.rs
```diff
@@ -106,6 +111,98 @@ fn value_to_string(value: &Value) -> String {
...
+fn deserialize_datetime<'de, D>(
+    deserializer: D,
+) -> Result<tv::DateTime, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    i64::deserialize(deserializer).map(tv::DateTime::from_timestamp_nanos)
+}
+
+/// An equivalent type to [`tantivy::schema::Value`] but uses tagging in its serialization to
```
adamreichold, line 138: I am not sure I understand why this requires a different serialization schema compared to the upstream `Value` type.

Cargo.toml
adamreichold: Sorry, I was unspecific. Why do you use a tilde requirement `~1.0` meaning `>=1.0, <1.1` instead of the default type `1.0` meaning `^1.0` meaning `>=1.0, <2.0`?

GodTamIt: I wanted to be more conservative/more explicit with the permissible versions, since unexpected changes in `serde` could make previously pickled types might not be compatible. But `1.0` (aka `^1.0`) should be fine, since breaking changes aren't likely with minor version changes.

src/document.rs
```diff
@@ -282,6 +379,67 @@ impl fmt::Debug for Document {
...
+}
+
+impl<'de> Deserialize<'de> for Document {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        struct MapVisitor;
+
+        impl<'de> Visitor<'de> for MapVisitor {
```
adamreichold, line 405: Wouldn't it be easier to just deserialize `BTreeMap<String, Vec<SerdeValue>>` and then convert this to `BTreeMap<String, Vec<Value>>`?

adamreichold: (I think the re-allocation could even be avoid if the conversion uses `.into_iter().map(..).collect()` as the types should layout compatible.)

src/document.rs
```diff
@@ -282,6 +379,67 @@ impl fmt::Debug for Document {
...
+impl Serialize for Document {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut map =
+            serializer.serialize_map(Some(self.field_values.len()))?;
+        for (k, v) in &self.field_values {
+            let ser_v: Vec<_> =
+                v.iter().cloned().map(SerdeValue::from).collect();
```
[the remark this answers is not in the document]
GodTamIt: These are great suggestions, thanks!

src/document.rs
GodTamIt: These are great suggestions, done.

src/schema.rs
```diff
@@ -25,4 +27,24 @@ impl Schema {
             _ => py.NotImplemented(),
         }
     }
+
+    #[staticmethod]
+    fn _internal_from_pythonized(serialized: &PyAny) -> PyResult<Self> {
```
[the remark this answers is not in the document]
adamreichold: Agreed. The social convention should suffice. Thanks for looking into it!

src/document.rs
```diff
@@ -282,6 +428,42 @@ impl fmt::Debug for Document {
...
+        map.end()
+    }
+}
+
+impl<'de> Deserialize<'de> for Document {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        HashMap::<String, Vec<SerdeValue>>::deserialize(deserializer).map(
```
adamreichold, line 452: I would recommend sticking to `BTreeMap` as the conversion from `HashMap` to `BTreeMap` will definitely need to re-allocate whereas `BTreeMap` could potentially avoid it.

GodTamIt: Oops, good catch, done.

adamreichold (approved):

GodTamIt:
Thanks for the reviews! This can land whenever

PR #99, opened 2023-07-21
Support copy, deepcopy, eq on types

GodTamIt:
This change implements the `__richcmp__`, `__copy__()` and `__deepcopy__()` functions for certain types that can be copied and compared.


cjrh:
Looks good. Just need python tests on the types with the new impls. Document, Facet and SearchResult right?

src/lib.rs
```diff
@@ -94,3 +94,51 @@ pub(crate) fn get_field(
...
+        impl $pyo3_ty {
+            pub fn __deepcopy__(&self, _memo: &::pyo3::types::PyDict) -> Self {
+                self.clone()
+            }
+        }
+    };
+}
+
+#[macro_export]
+macro_rules! impl_py_eq {
```
GodTamIt, line 123: I've added support for equality as well, since it makes it significantly easier to test and compare types.

Cargo.toml
```diff
@@ -22,4 +22,4 @@ serde_json = "1.0.91"
 
 [dependencies.pyo3]
 version = "0.18.0"
-features = ["chrono", "extension-module"]
+features = ["chrono", "extension-module", "multiple-pymethods"]
```
adamreichold, line 25: I would really suggest to avoid this feature if it is at all possible because it relies on a somewhat brittle backend depending on "life before main" to collect the impl blocks.

adamreichold: In this case, I think it would be possible to just move the macro invocations into the existing impl blocks.

src/document.rs
```diff
@@ -147,12 +150,16 @@ fn value_to_string(value: &Value) -> String {
 ///             {"unsigned": 1000, "signed": -5, "float": 0.4},
 ///             schema,
 ///         )
-#[pyclass]
-#[derive(Default)]
+#[pyclass(module = "tantivy")]
```
adamreichold, line 153: The module specification is necessary because of the usage of multiple impl blocks, right?

src/searcher.rs
```diff
@@ -194,13 +202,17 @@ impl Searcher {
...
-#[pyclass]
-#[derive(Clone, Debug)]
+#[pyclass(module = "tantivy")]
+#[derive(Clone, Debug, PartialEq)]
 pub(crate) struct DocAddress {
     pub(crate) segment_ord: tv::SegmentOrdinal,
     pub(crate) doc: tv::DocId,
 }
 
+impl_py_copy!(DocAddress);
```
adamreichold, line 212: I think rather than support copying, `DocAddress` should become a `#[pyclass(frozen)]` because it is immutable and hence references can be copied freely without any shared mutability issues, i.e. there should never be a need for deep copying.

src/schema.rs
```diff
@@ -1,14 +1,18 @@
+use crate::impl_py_eq;
 use pyo3::prelude::*;
 use tantivy as tv;
 
 /// Tantivy schema.
 ///
 /// The schema is very strict. To build the schema the `SchemaBuilder` class is
 /// provided.
-#[pyclass]
+#[pyclass(module = "tantivy")]
```
adamreichold, line 9: This should probably also be `frozen` because no method ever takes `&mut self`.

src/query.rs
```diff
@@ -18,4 +18,16 @@ impl Query {
     fn __repr__(&self) -> PyResult<String> {
         Ok(format!("Query({:?})", self.get()))
     }
+
+    pub fn __copy__(&self) -> Self {
+        Self {
+            inner: self.get().box_clone(),
```
adamreichold, line 24: Maybe implement the usual `Clone` trait manually and then use the macros? (But again, this class seems immutable and `#[pyclass(frozen)]` appears preferable to copy support.)

src/facet.rs
```diff
@@ -10,12 +11,16 @@ use tantivy::schema;
...
 /// and /electronics.
-#[pyclass]
-#[derive(Clone)]
+#[pyclass(module = "tantivy")]
+#[derive(Clone, PartialEq)]
 pub(crate) struct Facet {
     pub(crate) inner: schema::Facet,
 }
 
+impl_py_copy!(Facet);
```
adamreichold, line 20: Similarly, immutable and hence better to freeze than to copy.

Cargo.toml
cjrh: Great feedback thx @adamreichold 👍🏼

src/searcher.rs
cjrh: I didn't realise that was possible, this is much better yes.

Cargo.toml
GodTamIt: I tried this initially but it isn't too happy, see: https://github.com/PyO3/pyo3/issues/2862

We could avoid decide macros and just write the code manually?

GodTamIt:
I just realized that marking classes as `frozen` won't work with pickling 🙁

Cargo.toml
adamreichold: Ah yes, you are right, the macros will mess with the `#[pymethods]` proc macro. Sorry for suggesting it.

There are plans to avoid the boiler plate by adding syntax like `#[pyclass(eq, ord)]` to automatically derive the Python methods based on the Rust traits, but it isn't finished yet, c.f. https://github.com/PyO3/pyo3/pull/3195.

In summary, I think open-coding these short methods is the best approach for now. Thanks for making the extra effort.

adamreichold:
> I just realized that marking classes as `frozen` won't work with pickling slightly_frowning_face (`__setstate__` needs the class to be mutable)

Since immutability is highly desirable for both performance and correctness, especially in a language like Python built around pervasive shared mutability, I would recommend not dropping it easily to implement the Pickle protocol. (PyO3 is even planning to make frozen classes the default eventually due to this.)

Maybe we can use #96 to take a step back and reconsider why Pickle support was requested? Is full integration with the Pickle protocol necessary or only (de-)serialization from `PyBytes` which could also be implemented using methods returning newly created immutable instances?

GodTamIt:
> Since immutability is highly desirable for both performance and correctness, especially in a language like Python built around pervasive shared mutability, I would recommend not dropping it easily to implement the Pickle protocol. (PyO3 is even planning to make frozen classes the default eventually due to this.)

I was able to accomplish pickling with `__reduce__` and/or `__getnewargs__` and `pythonize` in the change. No mutability required for those.

This change is okay regardless.

src/document.rs
```diff
@@ -552,6 +554,24 @@ impl Document {
...
+    }
+
+    fn __richcmp__(&self, other: &Self, op: CompareOp) -> PyResult<bool> {
+        match op {
+            CompareOp::Eq => Ok(self == other),
+            CompareOp::Ne => Ok(self != other),
+            _ => Err(PyNotImplementedError::new_err(format!(
+                "{op:?} op is not supported"
+            ))),
+        }
```
cjrh, line 573: I believe the more expected thing is to return `NotImplemented` for unimplemented Ops.  This is discussed briefly in the PyO3 docs:

https://pyo3.rs/v0.18.3/class/object.html?highlight=alternatively#comparisons

image

adamreichold (approved):

PR #100, opened 2023-07-21
Expose IndexWriter::wait_merging_threads()

GodTamIt:
This can be useful if certain clients wants to wait for the merging to finish before issuing further commits.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -28,8 +28,34 @@ const RELOAD_POLICY: &str = "commit";
 /// on the index object.
 #[pyclass]
 pub(crate) struct IndexWriter {
-    inner_index_writer: tv::IndexWriter,
+    inner_index_writer: Option<tv::IndexWriter>,
     schema: tv::schema::Schema,
+}
+
+impl IndexWriter {
+    fn inner(&self) -> PyResult<&tv::IndexWriter> {
+        self.inner_index_writer.as_ref().ok_or_else(|| {
+            exceptions::PyValueError::new_err(
+                "IndexWriter was consumed and no longer in a valid state",
+            )
+        })
+    }
+
+    fn inner_mut(&mut self) -> PyResult<&mut tv::IndexWriter> {
+        self.inner_index_writer.as_mut().ok_or_else(|| {
+            exceptions::PyValueError::new_err(
+                "IndexWriter was consumed and no longer in a valid state",
+            )
+        })
+    }
+
+    fn take_inner(&mut self) -> PyResult<tv::IndexWriter> {
+        self.inner_index_writer.take().ok_or_else(|| {
+            exceptions::PyValueError::new_err(
+                "IndexWriter was consumed and no longer in a valid state",
+            )
+        })
+    }
 }
 
 #[pymethods]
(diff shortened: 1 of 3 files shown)


cjrh:
Looks good. Can you add a test (in python) that calls the new wait method?  Like a python version of one of these in the tantivy tests, e.g.

```rust
    #[test]
    fn test_with_merges() -> crate::Result<()> {
        let mut schema_builder = schema::Schema::builder();
        let text_field = schema_builder.add_text_field("text", TEXT);
        let index = Index::create_in_ram(schema_builder.build());
        let reader = index
            .reader_builder()
            .reload_policy(ReloadPolicy::Manual)
            .try_into()?;
        let num_docs_containing = |s: &str| {
            let term_a = Term::from_field_text(text_field, s);
            reader.searcher().doc_freq(&term_a).unwrap()
        };
        // writing the segment
        let mut index_writer = index.writer(12_000_000).unwrap();
        // create 8 segments with 100 tiny docs
        for _doc in 0..100 {
            index_writer.add_document(doc!(text_field=>"a"))?;
        }
        index_writer.commit()?;
        for _doc in 0..100 {
            index_writer.add_document(doc!(text_field=>"a"))?;
        }
        //  this should create 8 segments and trigger a merge.
        index_writer.commit()?;
        index_writer.wait_merging_threads()?;
        reader.reload()?;
        assert_eq!(num_docs_containing("a"), 200);
        assert!(index.searchable_segments()?.len() < 8);
        Ok(())
    }
```

Or if you have a simpler idea for a test that's fine too.

src/searcher.rs
```diff
@@ -164,6 +164,12 @@ impl Searcher {
         self.inner.num_docs()
     }
 
+    /// Returns the number of segments in the index.
+    #[getter]
+    fn num_segments(&self) -> usize {
```
GodTamIt, line 169: I did have to add this to expose the number of segments to the Python bindings to access it from my test.

It's not completely related but we do expose it already in `__repr__()` so I figured it may be useful anyways. If so desired, I can break this out in another change.

src/index.rs
```diff
@@ -28,10 +28,36 @@ const RELOAD_POLICY: &str = "commit";
...
 pub(crate) struct IndexWriter {
-    inner_index_writer: tv::IndexWriter,
+    inner_index_writer: Option<tv::IndexWriter>,
     schema: tv::schema::Schema,
 }
 
+impl IndexWriter {
+    fn inner(&self) -> PyResult<&tv::IndexWriter> {
+        self.inner_index_writer.as_ref().ok_or_else(|| {
+            exceptions::PyValueError::new_err(
```
adamreichold, line 38: I do wonder whether it makes sense to turn this into a `ValueError` since this happening would be a logic bug in the calling Python code, i.e. a `PanicException` seems appropriate?

adamreichold (approved):
Really not sure about which way to go with the exception, otherwise looks fine to me.

src/index.rs
cjrh: I'd have gone for something more python-native like [RuntimeError](https://docs.python.org/3/library/exceptions.html#RuntimeError) myself, but no strong feelings about this.

cjrh (approved):

GodTamIt:
I went with a `RuntimeError` there. Nothing fits perfectly here, but based on [the documentation](https://docs.python.org/3/library/exceptions.html#RuntimeError), `RuntimeError` seems to be the appropriate catch-all.

PR #101, opened 2023-07-23
Clean up document.rs

GodTamIt:
Custodial change with no functional difference to combine non-Python `impl Document` blocks together and move code so that all `impl Document` blocks are located next to each other. Makes it easier to read and find things when trying to change `Document`.


cjrh (approved):
Changes make sense to me 👍🏼

adamreichold (approved):

PR #105, opened 2023-08-04
Add support for booleans in schemas and docs

GodTamIt:
Previously, the bindings did not support adding boolean fields to either the schema or the document. This adds support for it.

### DIFF
--- a/src/schemabuilder.rs
+++ b/src/schemabuilder.rs
@@ -98,11 +98,12 @@ impl SchemaBuilder {
     ///         content of the field can be later restored from a Searcher.
     ///         Defaults to False.
     ///     indexed (bool, optional): If true sets the field to be indexed.
-    ///     fast (str, optional): Set the numeric options as a single-valued
-    ///         fast field. Fast fields are designed for random access. Access
-    ///         time are similar to a random lookup in an array. If more than
-    ///         one value is associated to a fast field, only the last one is
-    ///         kept.
+    ///     fast (str, optional): Set the numeric options as a fast field. A
+    ///         fast field is a column-oriented fashion storage for tantivy.
+    ///         It is designed for the fast random access of some document
+    ///         fields given a document id. Read access performance is
+    ///         comparable to that of an array lookup. If more than one value
+    ///         is associated with a fast field, only the last one is kept.
     ///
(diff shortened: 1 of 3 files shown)


src/schemabuilder.rs
```diff
@@ -196,6 +196,48 @@ impl SchemaBuilder {
...
 
+    /// Add a new boolean field to the schema.
+    ///
+    /// Args:
+    ///     name (str): The name of the field.
+    ///     stored (bool, optional): If true sets the field as stored, the
+    ///         content of the field can be later restored from a Searcher.
+    ///         Defaults to False.
+    ///     indexed (bool, optional): If true sets the field to be indexed.
+    ///     fast (str, optional): Set the bool options as a single-valued fast
```
adamreichold, line 207: This documentation does not look right for `fast: bool` and also for how this works since the bump to Tantivy 0.20.x. But admittedly this issues seem to apply to the other schema methods as well.

GodTamIt: I updated the documentation to reflect the 0.20.x. There were some other typos there that I just fixed up.

cjrh (approved):

src/schemabuilder.rs
adamreichold: I know this is not part of this change, but while we are discussing this:

* I think the sentence about "array access" is not correct any more because fast fields use a more complex encoding since 0.20.x which is slightly slower but much more flexible.
* IIRC, this flexibility primarily means that the cardinality limits are gone, i.e. all fast field can be optional and/or multi-valued and this like the best codec is determined automatically.

I think the best upstream documentation is currently https://docs.rs/tantivy/latest/tantivy/fastfield/index.html which we should probably reproduce and/or link here.

GodTamIt: I still have `Read access performance is comparable to that of an array lookup.` since that's still on the 0.20.x documentation you linked.

https://docs.rs/tantivy/latest/tantivy/schema/struct.NumericOptions.html#method.set_fast seems to still have this language - is that out of date?

GodTamIt:
Thanks! All changes are done, so this is ready to land whenever :)

cjrh:
Thanks @GodTamIt 👍🏼

PR #114, opened 2023-08-15
Support lenient parser

### LINKED ISSUE #109: Support lenient parser in bindings
Now that https://github.com/quickwit-oss/tantivy/pull/2129 has landed, it'd be nice to start thinking about lenient parsing support in the bindings. Obviously, a new version of the crate has to be released but starting to think about this.

I've started working on this, but there are a few points I'd like to iron out how we'd expose the `QueryParserError` enum. It'd be nice to have types for the errors, but `QueryParserError` is currently a Rust enum.

An (strawman) idea I had was to have separate classes for each error type. Then, optionally, organize them in a submodule underneath `tantivy`, i.e. `tantivy.query_parser_error` or something. Then, the interface could look something like this from the Python side:

```python
@dataclass
class ParseQueryLenientResult:
    query: Query
    errors: List[Union[query_parser_error.SyntaxError, query_parser_error.UnsupportedQuery, ...]]

def parse_query_lenient(query: str, default_field_names: Optional[List[str]]) -> ParseQueryLenientResult:
    ...
```

cc @adamreichold @cjrh if y'all have thoughts

GodTamIt:
**Note:** this is a strawman for supporting lenient parser in the bindings. Of course, this can't be landed until there is a new release of the `tantivy` crate with the lenient parser included.

Closes #109

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,9 +15,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.1"
+version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ea5d730647d4fadd988536d06fecce94b7b4f2a7efdae548f1cf4b63205518ab"
+checksum = "43f6cb1bf222025340178f382c426f13757b2960e89779dfcb319c32542a5a41"
 dependencies = [
  "memchr",
 ]
@@ -27,12 +27,6 @@ name = "allocator-api2"
 version = "0.2.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0942ffc6dcaadf03badf6e6a2d0228460359d5e34b57ccdc720b7382dfbd5ec5"
-
-[[package]]
-name = "android-tzdata"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
 
 [[package]]
 name = "android_system_properties"
@@ -51,13 +45,13 @@ checksum = "bddcadddf5e9015d310179a59bb28c4d4b9920ad0f11e8e14dbadf654890c9a6"
 
 [[package]]
 name = "async-trait"
-version = "0.1.73"
+version = "0.1.64"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bc00ceb34980c03614e35a3a4e218276a0a824e911d07651cd0d858a51e8c0f0"
+checksum = "1cd7fce9ba8c3c042128ce72d8b2ddbf3a05747efb67ea0313c635e10bda47a2"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.37",
+ "syn 1.0.107",
 ]
 
 [[package]]
@@ -68,9 +62,9 @@ checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
 name = "base64"
-version = "0.21.4"
+version = "0.21.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9ba43ea6f343b788c8764558649e08df62f86c6ef251fdaeb1ffd010a9ae50a2"
+checksum = "604178f6c5c21f02dc555784810edfb88d34ac2c73b2eae109655649ee73ce3d"
 
 [[package]]
 name = "bitflags"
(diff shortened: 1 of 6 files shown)


src/parser_error.rs
```diff
@@ -0,0 +1,972 @@
...
+    num::{IntErrorKind, ParseFloatError, ParseIntError},
+    str::ParseBoolError,
+};
+
+use pyo3::prelude::*;
+use tantivy::{self as tv, schema::FacetParseError};
+
+// TODO(https://github.com/PyO3/pyo3/issues/1190): Expose this to bindings once trait <-> ABC is
+// supported in PyO3.
+pub(crate) trait QueryParserError {
```
GodTamIt, line 13: This isn't a base class because it seems base classes make it tough to do things like `IntoPy` and upcasting.

See: https://github.com/PyO3/pyo3/issues/1836 and https://github.com/PyO3/pyo3/issues/787

src/searcher.rs
```diff
@@ -129,7 +129,7 @@ impl Searcher {
             if let Some(order_by) = order_by_field {
                 let collector = TopDocs::with_limit(limit)
                     .and_offset(offset)
-                    .order_by_u64_field(order_by);
+                    .order_by_fast_field(order_by, tv::Order::Desc);
```
GodTamIt, line 132: This change is necessary to compile against ToT of `tantivy`.

Cargo.toml
```diff
@@ -14,8 +14,9 @@ crate-type = ["cdylib"]
 pyo3-build-config = "0.19.1"
 
 [dependencies]
+base64 = "0.21"
 chrono = "0.4.23"
-tantivy = "0.20.1"
+tantivy = { git = "https://github.com/quickwit-oss/tantivy.git" }
```
GodTamIt, line 19: This is because there hasn't been a release with lenient parser support in `tantivy` yet.

Cargo.toml
```diff
@@ -14,8 +14,9 @@ crate-type = ["cdylib"]
 pyo3-build-config = "0.19.1"
 
 [dependencies]
+base64 = "0.21"
```
GodTamIt, line 17: This is necessary to access the `base64` error that `tantivy` holds. This dependency won't change much otherwise, since it's also being used by `tantivy` anyways.

GodTamIt:
Ready for review now

r? @cjrh @adamreichold

cjrh:
Hoping to get around to it this weekend.

Cargo.lock
```diff
@@ -15,9 +15,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.0.2"
+version = "1.1.1"
```
cjrh, line 18: Many transitive dependencies are also getting updated in this PR. It makes me a little nervous but I am ok to continue with fixing this up. In future though, it would be better if we can restrict the dep changes to only what is required by the other changes in the same PR.

src/index.rs
```diff
@@ -399,6 +400,67 @@ impl Index {
...
+                    )));
+                }
+            }
+        } else {
+            for (field, field_entry) in self.index.schema().fields() {
+                if field_entry.is_indexed() {
+                    default_fields.push(field);
+                }
+            }
+        }
```
cjrh, line 452: Out of mainly curiosity, I looked at removing the mutability of `default_fields` by returning the vec from the if-expression:

```rust
    pub fn parse_query_lenient(
        &self,
        query: &str,
        default_field_names: Option<Vec<String>>,
    ) -> PyResult<(Query, Vec<PyObject>)> {
        let schema = self.index.schema();

        let default_fields = if let Some(default_field_names_vec) =
            default_field_names
        {
            default_field_names_vec
                .iter()
                .map(|field_name| {
                    schema
                        .get_field(field_name)
                        .map_err(|_err| {
                            exceptions::PyValueError::new_err(format!(
                                "Field `{field_name}` is not defined in the schema."
                            ))
                        })
                        .and_then(|field| {
                            schema.get_field_entry(field).is_indexed().then_some(field).ok_or(
                                exceptions::PyValueError::new_err(
                                    format!(
                                        "Field `{field_name}` is not set as indexed in the schema."
                                    ),
                                ))
                        })
                }).collect::<Result<Vec<_>, _>>()?
        } else {
            self.index
                .schema()
                .fields()
                .filter_map(|(f, fe)| fe.is_indexed().then_some(f))
                .collect::<Vec<_>>()
        };

```

I don't find it super compelling to make this change though. What do you think? I'm going to approve the PR regardless.

cjrh (approved):

src/index.rs
GodTamIt: Why not. I have a preference for functional-styling here.

Cargo.lock
GodTamIt: I've undone these changes now.

wallies:
@GodTamIt you should be good to merge this and ill cut a new release, as the lenient parser was in tantivy 0.21

PR #132, opened 2023-09-26
Update to tantivy v0.21.0

GodTamIt:
This change bumps the version of `tantivy` to v0.21.0.

Notably in this change is the addition of `Order` to specify order_by sort order for fast fields that was introduced in the new version.

### DIFF
--- a/src/searcher.rs
+++ b/src/searcher.rs
@@ -41,6 +41,22 @@ impl ToPyObject for Fruit {
 }
 
 #[pyclass(frozen, module = "tantivy")]
+#[derive(Clone, Copy, Deserialize, PartialEq, Serialize)]
+pub(crate) enum Order {
+    Asc,
+    Desc,
+}
+
+impl From<Order> for tv::Order {
+    fn from(order: Order) -> Self {
+        match order {
+            Order::Asc => tv::Order::Asc,
+            Order::Desc => tv::Order::Desc,
+        }
+    }
+}
+
+#[pyclass(frozen, module = "tantivy")]
 #[derive(Clone, Default, Deserialize, PartialEq, Serialize)]
 /// Object holding a results successful search.
 pub(crate) struct SearchResult {
@@ -126,11 +142,13 @@ impl Searcher {
     ///         fields.
     ///     offset (Field, optional): The offset from which the results have
     ///         to be returned.
+    ///     order_by_order (Order, optional): The order in which the results
+    ///         should be sorted. If not specified, defaults to descending.
     ///
     /// Returns `SearchResult` object.
     ///
     /// Raises a ValueError if there was an error with the search.
-    #[pyo3(signature = (query, limit = 10, count = true, order_by_field = None, offset = 0))]
+    #[pyo3(signature = (query, limit = 10, count = true, order_by_field = None, offset = 0, order_by_order = Order::Desc))]
     fn search(
         &self,
         py: Python,
@@ -139,6 +157,7 @@ impl Searcher {
         count: bool,
         order_by_field: Option<&str>,
         offset: usize,
+        order_by_order: Order,
     ) -> PyResult<SearchResult> {
         py.allow_threads(move || {
             let mut multicollector = MultiCollector::new();
@@ -153,7 +172,7 @@ impl Searcher {
                 if let Some(order_by) = order_by_field {
                     let collector = TopDocs::with_limit(limit)
                         .and_offset(offset)
-                        .order_by_u64_field(order_by);
+                        .order_by_fast_field(order_by, order_by_order.into());
                     let top_docs_handle =
                         multicollector.add_collector(collector);
                     let ret = self.inner.search(query.get(), &multicollector);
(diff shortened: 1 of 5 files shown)


src/searcher.rs
```diff
@@ -139,6 +157,7 @@ impl Searcher {
         count: bool,
         order_by_field: Option<&str>,
         offset: usize,
+        order_by_order: Order,
```
cjrh, line 160: How strongly do you feel that this should be called `order_by_order` instead of just `order`?

src/searcher.rs
```diff
@@ -126,11 +142,13 @@ impl Searcher {
...
     ///     offset (Field, optional): The offset from which the results have
     ///         to be returned.
+    ///     order_by_order (Order, optional): The order in which the results
+    ///         should be sorted. If not specified, defaults to descending.
     ///
     /// Returns `SearchResult` object.
     ///
     /// Raises a ValueError if there was an error with the search.
-    #[pyo3(signature = (query, limit = 10, count = true, order_by_field = None, offset = 0))]
+    #[pyo3(signature = (query, limit = 10, count = true, order_by_field = None, offset = 0, order_by_order = Order::Desc))]
```
cjrh, line 151: Just checking my understanding: this must be `Desc` because the function it replaces, `order_by_u64_field` was also descending, according to these docs: https://docs.rs/tantivy/0.19.2/tantivy/collector/struct.TopDocs.html#method.order_by_u64_field

Is that correct?

src/searcher.rs
GodTamIt: Not particularly strongly here. Just thought it'd be more specific about which other parameter its associated with but I don't anticipate another ordering field any time soon?

tests/tantivy_test.py
```diff
@@ -555,10 +571,10 @@ def test_create_readers(self):
 
 class TestSearcher(object):
     def test_searcher_repr(self, ram_index, ram_index_numeric_fields):
-        assert repr(ram_index.searcher()) == "Searcher(num_docs=3, num_segments=1)"
+        assert repr(ram_index.searcher()) == "Searcher(num_docs=3, num_segments=3)"
```
cjrh, line 574: Hmm, why has this changed back?

src/searcher.rs
GodTamIt: Yep, that's correct. It allows for code compatibility with the new version, which I defaulted to since there didn't seem to otherwise be a good reason to change the default.

tests/tantivy_test.py
cjrh: The writer should have been created with a large enough memory limit to allow 1 segment:

```python
writer = index.writer(10_000_000, 1)
```

^^ This is the writer inside the `create_index()` that is being called from `ram_index`.  I wonder if this change is related to the tantivy upgrade itself?

cjrh:
Loving your work Chris 👍🏼

src/searcher.rs
cjrh: I can live with `order_by_order`, but it feels "unpythonlike" 😄 .  We can hear from the other maintainers how they feel.

tests/tantivy_test.py
cjrh: @GodTamIt Could you test whether increasing that 10M brings the `num_segments` back down to 1?

GodTamIt: Looks like a small behavior change. Increasing the memory limit seems to have fixed the issue.

adamreichold: The memory accounting in Tantivy was improved and now tracks memory that was not included before, hence the same workload apparently needs more memory. (Actually, it needs the same, but the limit now includes more of the actual working set.)

src/searcher.rs
adamreichold: To add yet another name to consider, how about `order_in_direction` (and matching `enum Direction`)?

src/searcher.rs
```diff
@@ -40,6 +40,22 @@ impl ToPyObject for Fruit {
     }
 }
 
+#[pyclass(frozen, module = "tantivy")]
+#[derive(Clone, Copy, Deserialize, PartialEq, Serialize)]
+pub(crate) enum Order {
+    Asc,
```
adamreichold, line 46: Could we add doc comments to spell these out as `Ascending` and `Descending` as non-native speakers might have trouble recognizing the abbreviations.

GodTamIt: I added doc comments above each discriminant. Let me know if that works

src/searcher.rs
GodTamIt: I went with `order` (and a matching `enum Order`). I have a strong preference for an API surface and naming convention that closely mimics the original `tantivy` APIs, since I often go to those docs to see how things work in these bindings (as an outside user) :)

cjrh: Same. I wish that `order_by_field` was just called `order_by`, but we can't easily change that now probably.

cjrh (approved):
I'm happy but I'd like to wait for @adamreichold to approve also.

adamreichold (approved):

PR #138, opened 2023-10-09
bump package version to 0.21.0

wallies:


wallies:
@adamreichold @GodTamIt @cjrh @Sidhant29 Be good to get your perspective. Do we need to look at this naming scheme e.g. 0.21.0 tantivy version with an extra .x for tantivy-py updates or do we not need to solve for this

adamreichold:
While having the versions of the `tantivy` crate and the `tantiv-py` package agree is nice, I consider a bit of coincidence and hence would not go against the established semver practice of using `major.minor.patch` versions just to accommodate it.

cjrh:
@adamreichold I disagree. The long history of the APSW package, and my user experience of working with that package over many years, has made it quite clear that _for wrappers_, having a simple correspondence between the inner package (SQLite) and the wrapper package (APSW) is valuable.

They describe it here: https://rogerbinns.github.io/apsw/about.html#apsw-and-sqlite-versions

It may not be strict semver but semver doesn't have a good solution for wrapper packages anyway.

cjrh:
Here, recently, the arrow-datafusion-python wrapper discussed the same problem: https://github.com/apache/arrow-datafusion-python/issues/123

Here, several years ago, a comment at semver saying that semver is not intended to describe package dependences. However, they offer the idea of, e.g. `M1.m1.p1+tantivy.M2.m2.p2` as a way of representing both (M1 for tantivy-py major version, M2 for tantivy major version).  It looks terrible though :)

GodTamIt:
This might be a bit of a bikeshed but I would say that having `tantivy` and `tantivy-py` version numbers match would be my preference. Chief among the reasons is that it makes tracking bugs and version history significantly easier.

However, I don't have strong opinions on how we'd differentiate between builds of `tantivy-py` with the same underlying `tantivy` version.

cjrh:
Ok @wallies let's move forward with semver.

cjrh (approved):

PR #139, opened 2023-10-09
chore: deprecate python 3.7 and add 3.12

wallies:


adamreichold (commented):
I think the first version of PyO3 to officially support Python 3.12 will be 0.20 which is not released yet, so I'd suggest waiting to make that bump before committing. Otherwise we might run into too many problems which originate in PyO3-CPython incompatibilities.

cjrh:
We should be able to have 3.12 running in CI though?  I agree that we should not make a release of tantivy-py that supports 3.12 until PyO3 officially supports it, that's a good point 👍🏼

cjrh (approved):

PR #150, opened 2023-10-31
query & geocoding

jberthier-ma:


cjrh:
Hi @jberthier-ma, thanks for the contribution 😄 

Would you be able to add tests?

The other thing I was wondering was the difference in output for the query `__repr__`. Could you give an example of the before and after? I am thinking through whether this is a semver break, since it modifies an existing user-facing output, so I want to know whether the change in `__repr__` output is worth it.

PR #154, opened 2023-11-18
Change to action `dtolnay/rust-toolchain` and 1.73

cjrh:
This changes to the `dtolnay/rust-toolchain` action because `actions-rs` [is deprecated](https://github.com/actions-rs/toolchain/issues/216). It also pins the rust compiler to 1.73.0 because of an [internal compiler error in 1.74](https://github.com/rust-lang/rust/issues/117976).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,10 +28,9 @@ jobs:
         uses: actions/checkout@61b9e3751b92087fd0b06925ba6dd6314e06f089
 
       - name: Rust toolchain
-        uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af
+        uses: dtolnay/rust-toolchain@1.73.0
         with:
-          toolchain: stable
-          components: rustfmt
+            components: rustfmt
 
       - name: Check Formatting
         run: cargo fmt --check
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -28,10 +28,9 @@ jobs:
         uses: actions/checkout@61b9e3751b92087fd0b06925ba6dd6314e06f089
 
       - name: Rust toolchain
-        uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af
+        uses: dtolnay/rust-toolchain@1.73.0
```
wallies, line 31: Can we use the hash like the other actions

cjrh: not sure? https://github.com/marketplace/actions/rustup-toolchain-install

cjrh: Maybe something like this?

```yaml
     # Installs the stable toolchain which preceded the most recent one by
     # the specified number of minor versions.
  - uses: dtolnay/rust-toolchain@439cf607258077187679211f12aa6f19af4a0af7
    with:
      toolchain: stable minus 8 releases
```

cjrh: Done

wallies (approved):

PR #157, opened 2023-11-20
fix ci allowlist

wallies:


.github/workflows/publish.yaml
```diff
@@ -103,23 +105,15 @@ jobs:
...
           command: build
           args: --release -o dist
 
-      - name: Build wheels - universal2
-        uses: messense/maturin-action@683eb33e2cda131da55757dbac7628b2ef22df2f
-        env:
-          PYO3_PYTHON: python${{ matrix.python-version }}
-        with:
-          command: build
-          args: --release -o dist --universal2
```
adamreichold, line 121: Just to make sure, the `--universal2` flag is not necessary when `target` is passed to the action?

.github/workflows/publish.yaml
```diff
@@ -36,20 +36,21 @@ jobs:
             files.pythonhosted.org:443
             ghcr.io:443
             pkg-containers.githubusercontent.com:443
+            cdn02.quay.io:443
```
adamreichold, line 39: I'd suggest moving to the other quay.io endpoint above. I think genereally, that sorting these based on the domains would be helpful.

wallies: Good idea, will sort them. I didn't have a choice of domains. The manylinux container downloads from quay.io registry and I suspect maturin uses the cdn. I'll make sure to add all the cdns for quay.io

cjrh (approved):

PR #158, opened 2023-11-20
Support passing dict to add JSON field

GodTamIt:
This allows you to now do the following:

```python
doc.add_json(
    "attributes",
    {
        "order": 1.2,
        "target": "submit-button",
        "cart": {"product_id": 133},
        "description": "das keyboard",
    },
)
```

It also fixes a panic that occurs when passing valid JSON that is not a map, originating [here](https://github.com/quickwit-oss/tantivy/blob/49448b31c69ceec4cbab0f68c38f608b0dbd1a3b/src/schema/value.rs#L312)

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -672,19 +672,28 @@ impl Document {
         self.add_value(field_name, bytes);
     }
 
-    /// Add a bytes value to the document.
+    /// Add a JSON value to the document.
     ///
     /// Args:
     ///     field_name (str): The field for which we are adding the bytes.
-    ///     value (str): The json object that will be added to the document.
+    ///     value (str | Dict[str, Any]): The JSON object that will be added
+    ///         to the document.
     ///
-    /// Raises a ValueError if the json is invalid.
-    fn add_json(&mut self, field_name: String, json: &str) -> PyResult<()> {
-        let json_object: serde_json::Value =
-            serde_json::from_str(json).map_err(to_pyerr)?;
-        self.add_value(field_name, json_object);
+    /// Raises a ValueError if the JSON is invalid.
+    fn add_json(&mut self, field_name: String, value: &PyAny) -> PyResult<()> {
+        type JsonMap = serde_json::Map<String, serde_json::Value>;
 
-        Ok(())
+        if let Ok(json_str) = value.extract::<&str>() {
+            let json_map: JsonMap =
+                serde_json::from_str(json_str).map_err(to_pyerr)?;
+            self.add_value(field_name, json_map);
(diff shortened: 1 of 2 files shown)


src/document.rs
```diff
@@ -676,15 +676,24 @@ impl Document {
...
+            let json_map: JsonMap =
+                serde_json::from_str(json_str).map_err(to_pyerr)?;
+            self.add_value(field_name, json_map);
+            Ok(())
+        } else if let Ok(json_map) = pythonize::depythonize::<JsonMap>(value) {
+            self.add_value(field_name, json_map);
+            Ok(())
+        } else {
+            Err(to_pyerr("Invalid JSON object. Expected valid JSON string or Dict[str, Any]."))
+        }
```
cjrh, line 696: I personally would prefer a separate method for the additional type for two reasons:
1. JSON is a string type, so it would be jarring to pass a dict
2. The ability to add multiple types is more discoverable if there are different methods for them, because LSP/intellisense shows the options when you type `blah.add_|`.

It's possible other maintainers might feel different though. For example (this probably won't compile, I'm just sketching it out):

```suggestion
    // As before
    fn add_json(&mut self, field_name: String, json: &str) -> PyResult<()> {
        let json_object: serde_json::Value =
            serde_json::from_str(json).map_err(to_pyerr)?;
        self.add_value(field_name, json_object);
        Ok(())
    }
    
    fn add_dict(&mut self, field_name: String, value: &PyDict) -> PyResult<()> {
        type JsonMap = serde_json::Map<String, serde_json::Value>;
        if let Ok(json_map) = pythonize::depythonize::<JsonMap>(value) {
            self.add_value(field_name, json_map);
            Ok(())
        } else {
            Err(to_pyerr("Invalid JSON object. Expected a dict[str, Any]."))
        }
```

Offtopic, I suppose if we wanted to restrict the `PyDict` above to `dict[str, Any]`, we'd probably have to use a stubs file? Is there a way to express those type hints in any PyO3 features?

cjrh (changes requested):
This is great, I think it will be quite useful to be able to work with dicts directly.  I'd prefer json and dict to be handled in separate methods though.

What do you think?

src/document.rs
cjrh: I think I was wrong about this. Your original PR was better, because the JSON being referred to is the tantivy field type, not the python incoming parameter type 🙃 

If you change it back to what you originally had, I'll merge it.  Sorry about this 🙏🏼

cjrh (approved):
Thanks!

PR #162, opened 2023-11-21
Upgrade 0.21.1 and merge upstream

wallies:


wallies:
i have no idea what happened here, i was pulling the specific upstream commits over. I must have messed one up

PR #166, opened 2023-12-08
Added a failing test for ordering on date field

pratyushmittal:
The recent changes seems to have broken the `order_by_field` option for `date` fields. It gives the error `'Field "order" is of type U64!=Date'` when sorting on date field.

I have added the tests to reproduce the above.

Sorting on date field worked fine until `0.17`. 

I also checked the change log in `tantivy`. The `v0.21` specifically mentions:

> Fix a regression from 0.20 where sort index by date wasn't working anymore https://github.com/quickwit-oss/tantivy/issues/2124(@PSeitz)

Hence I believe it should be working fine here too.

### DIFF
--- a/tests/tantivy_test.py
+++ b/tests/tantivy_test.py
@@ -384,6 +384,57 @@ class TestClass(object):
         searched_doc = index.searcher().doc(doc_address)
         assert searched_doc["title"] == ["Test title"]
 
+    def test_order_by_search_date(self):
+        schema = (
+            SchemaBuilder()
+            .add_date_field("order", fast=True)
+            .add_text_field("title", stored=True)
+            .build()
+        )
+
+        index = Index(schema)
+        writer = index.writer()
+
+        doc = Document()
+        doc.add_date("order", datetime.datetime(2020, 1, 1))
+        doc.add_text("title", "Test title")
+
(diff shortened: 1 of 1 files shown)


cjrh:
Thanks for looking into this.  One of us will have a closer look when we have time.

ditsuke:
It appears that `tantivy-py` has been using `order_by_fast_field` since updating to v0.21.0 ([11f8bc](https://github.com/quickwit-oss/tantivy-py/commit/11f8bc4611de663fbc34109388710da55d8ccb7d#diff-49a95f762fea0977e895625649db32c1cbf229a3ea4073cd0c73c2c46312f562L156-R179)), which is generic on the fast field value for `order_by_field`. This won't work for non-u64 fields unless the context is type-aware through some other means. The simplest bc solution is to revert to using `order_by_u64_field` which does a lax field_type => u64 coercion, but it needs tantivy core to expose it as `pub` again.

https://github.com/quickwit-oss/tantivy-py/blob/539129154160e850d3e7d9a8ec5bc358b9b121d7/src/searcher.rs?plain=1#L176-L179

cjrh:
The fix, along with your reproducer test here have been applied in #262. Thanks for reporting this 😄

PR #169, opened 2023-12-10
Add all Python stubs in tantivy.pyi

mocobeta:
This is the follow-up of #168.

I added all stubs for currently exposed classes/public methods in tantivy-py. 
It seems this works fine for me - I checked that with VSCode.


cjrh (approved):

cjrh:
Thanks!

mocobeta:
@cjrh Thank you for the quick responce!

Issue #20, opened 2020-03-12
Support for other query method


theblackcat102:
Will there be support for other query method such as PhraseQuery, FuzzyTermQuery, RangeQuery?

poljar:
I don't have any plans to work on this myself, at least not for now.

Contributions of course are always welcome.

GuillaumePressiat:
Hi, 

How to contribute then ?
I doesn't see python code in the repo.

fuzzy matching absence is a no go for using tantivy in python even though tantivy seems very powerful

Many thanks,
Guillaume

mocobeta:
Hi guys! 
I'd like to contribute to this issue. For a good starter, I've already opened #175. ;) Could anyone review it?

cjrh:
#175 has been merged. I think either we should have a checklist of remaining query methods on this issue, or we create new separate enhancement issues for each of the remaining ones. Which is better?

mocobeta:
There are several Query classes in Tantivy.
https://docs.rs/tantivy/latest/tantivy/query/trait.Query.html#impl-Query-for-Box%3Cdyn+Query%3E

I think having a checklist is a great idea on this issue.

mocobeta:
I made a simple checklist to support all Tantivy queries (I don't have permission to edit the issue description.)

- [ ] AllQuery
- [ ] BooleanQuery
- [ ] BoostQuery
- [ ] ConstScoreQuery
- [ ] DisjunctionMaxQuery
- [ ] EmptyQuery
- [ ] FuzzyTermQuery
- [ ] MoreLikeThisQuery
- [ ] PhrasePrefixQuery
- [ ] PhraseQuery
- [ ] RangeQuery
- [ ] RegexQuery
- [x] TermQuery
- [ ] TermSetQuery

imaurer:
Looks like FuzzyTermQuery has been implemented, but not available in the current release on PyPI. Curious if this project has a cadence for releases. Thanks!

https://github.com/quickwit-oss/tantivy-py/commit/e95a4569d488059ce01f0cea4f1bc4a371f903a0

cjrh:
Hi @imaurer, we don't have a set cadence of releases unfortunately.  @wallies does our releases, I'm sure he'll get around to it within the next week or two.

aecio:
I just submitted a PR that exposes  DisjunctionMaxQuery. Hopefully, someone can review and it can make it to the next release as well.

BTW, thanks for this great library!

[...]

PR #175, opened 2023-12-17
Expose tantivy's TermQuery

mocobeta:
This may relate to #20.

This exposes Tantivy's `TermQuery` to tantivy-py.
I also noticed #53, but it seems incomplete and stale for a while.

- Add `Query.new_term_query()` method with proper type hints.
- Add a test method for TermQuery.

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -1,4 +1,5 @@
-use pyo3::prelude::*;
+use crate::{make_term, Schema};
+use pyo3::{exceptions, prelude::*, types::PyAny};
 use tantivy as tv;
 
 /// Tantivy's Query
@@ -18,4 +19,37 @@ impl Query {
     fn __repr__(&self) -> PyResult<String> {
         Ok(format!("Query({:?})", self.get()))
     }
+
+    /// Construct a Tantivy's TermQuery
+    #[staticmethod]
+    #[pyo3(signature = (schema, field_name, field_value, index_option = "position"))]
+    pub(crate) fn new_term_query(
+        schema: &Schema,
+        field_name: &str,
+        field_value: &PyAny,
+        index_option: &str,
+    ) -> PyResult<Query> {
+        make_term_query(schema, field_name, field_value, index_option)
+    }
+}
+
+fn make_term_query(
+    schema: &Schema,
+    field_name: &str,
+    field_value: &PyAny,
+    index_option: &str,
+) -> PyResult<Query> {
+    let term = make_term(&schema.inner, field_name, field_value)?;
+    let index_option = match index_option {
+        "position" => tv::schema::IndexRecordOption::WithFreqsAndPositions,
+        "freq" => tv::schema::IndexRecordOption::WithFreqs,
+        "basic" => tv::schema::IndexRecordOption::Basic,
+        _ => return Err(exceptions::PyValueError::new_err(
+            "Invalid index option, valid choices are: 'basic', 'freq' and 'position'"
+        ))
+    };
+    let inner = tv::query::TermQuery::new(term, index_option);
+    Ok(Query {
+        inner: Box::new(inner),
+    })
 }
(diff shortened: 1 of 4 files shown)


src/query.rs
```diff
@@ -18,4 +19,37 @@ impl Query {
...
+    /// Construct a Tantivy's TermQuery
+    #[staticmethod]
+    #[pyo3(signature = (schema, field_name, field_value, index_option = "position"))]
+    pub(crate) fn new_term_query(
+        schema: &Schema,
+        field_name: &str,
+        field_value: &PyAny,
+        index_option: &str,
+    ) -> PyResult<Query> {
+        make_term_query(schema, field_name, field_value, index_option)
```
adamreichold, line 32: Why the separate function to implement this instead of directly here in the method?

src/query.rs
```diff
@@ -18,4 +19,37 @@ impl Query {
     fn __repr__(&self) -> PyResult<String> {
         Ok(format!("Query({:?})", self.get()))
     }
+
+    /// Construct a Tantivy's TermQuery
+    #[staticmethod]
+    #[pyo3(signature = (schema, field_name, field_value, index_option = "position"))]
+    pub(crate) fn new_term_query(
```
adamreichold, line 26: I would suggest just `term_query`.

tests/tantivy_test.py
```diff
@@ -925,3 +925,15 @@ def test_document_snippet(self, dir_index):
             assert first.end == 23
             html_snippet = snippet.to_html()
             assert html_snippet == "The Old Man and the <b>Sea</b>"
+
+
+class TestQuery(object):
+    def test_term_query(self, ram_index):
+        index = ram_index
+        query = Query.new_term_query(index.schema, "title", "sea")
```
adamreichold, line 933: I think exposing the fundamental query types is not too useful without having combinators, i.e. we should at least add `union` and `intersection` combinators to combine queries on the Python side.

src/query.rs
mocobeta: Thanks for the suggestion. `term_query` sounds more concise.

src/query.rs
mocobeta: It's just for readability. I suppose we will have several factory methods like this for other query types.
But having all implementations here is also fine.

tests/tantivy_test.py
mocobeta: I guess you mean BooleanQuery by "union" and "intersection"? Exposing TermQuery itself may not help for real use cases. However, I'd like to take baby steps and not bloat this PR. 
We can support all other queries in a step-by-step manner once this is successfully merged.

adamreichold: Yes, they would be implemented via `BooleanQuery`, but mirror its API, there would most likely be `union` and `intersection` methods on `Query`.

Fair enough on doing this in a follow-up PR.

mocobeta: Sure, we should be able to have some fluent APIs to construct a bool query. It'll be a future work; anyway, thanks for your suggestion!

src/query.rs
mocobeta: Renamed in https://github.com/quickwit-oss/tantivy-py/pull/175/commits/e02b47986fc9c6f1de836ec49a9922a83e6ffd85.

src/query.rs
mocobeta: I moved the implementation to `term_query()` instead of using a utility function. https://github.com/quickwit-oss/tantivy-py/pull/175/commits/68bc4668e5f86cdcb47bdcfcac31c2812bd1d50f

adamreichold (approved):

mocobeta:
Hi,
This has received one approval (thank you @adamreichold for reviewing!)
I don't have permission to merge this into the main branch, could anyone take care of it?

mocobeta:
Thanks @wallies !

Issue #174, opened 2023-12-16
Add API reference to documentation site


mocobeta:
Now we have `.pyi` file (#167), we can generate API docs by [pdoc](https://pdoc.dev/).

For example, the following commands generate static HTML files in `apidoc` dir if you have `tantivy` module in your python path.

```
pip install pdoc
pdoc tantivy -o ./apidocs
```

Generated HTML looks very neat because [PyO3 exposes comments in `.rs` files as docstrings of tantivy module](https://pyo3.rs/v0.9.2/module#documentation), then pdoc collects all of them!

![tantivy_apidoc](https://github.com/quickwit-oss/tantivy-py/assets/1825333/a0302e27-07ae-47e3-aed2-85a18e84dd1b)

Perhaps we can publish the API doc to https://tantivy-py.readthedocs.io/en/latest/.
I'm not familiar with the CI pipeline to publish the docs site. Is there any suggestions about that?

cjrh:
The current workflow is I log into readthedocs on my own personal account and click "sync". I'm not happy with this, but I don't currently have enough permissions on this github repo to get the webhook working. Once the webhook is setup, docs deployment will be automatic.

fmassot:
@cjrh do you know which permission do you need?

cjrh:
@fmassot I think either I need "write" access, or I can add you or another admin to the readthedocs site and you can generate the integration there. Long term that is probably best anyway, so that I am not the only one with access to the readthedocs site.

- Documentation on readthedocs about the integration for github->readthedocs: https://docs.readthedocs.io/en/stable/guides/setup/git-repo-automatic.html
- Project page for tantivy-py on readthedocs: https://readthedocs.org/projects/tantivy-py/

This is the page where the integration can be performed:

![image](https://github.com/quickwit-oss/tantivy-py/assets/480395/7d129b05-102a-412e-bfdb-1f4f81108283)

When I click "resync" it fails:

![image](https://github.com/quickwit-oss/tantivy-py/assets/480395/6eab392c-3696-4cb2-bb6a-1e36e9911ea2)

If you create a login on readthedocs (make sure to use your github OAuth login), I can add your account to the tantivy-py project on readthedocs, and then you will be able to successfully click that `Resync webhook` button, which should work because it'll use the permissions on your github OAuth token.  Then I won't need any permissions change on tantivy-py.

Just let me know your login name on readthedocs and I'll add it to the project.

cjrh:
[...]

PR #182, opened 2023-12-31
Docs: incorporate pdoc to generate API docs

cjrh:
Incorporating the suggestion in #174.

This is a little messy.  Based on the [pdoc example for Mkdocs](https://github.com/mitmproxy/pdoc/tree/main/examples/mkdocs), it seems you have to run a script to execute pdoc before the mkdocs pass. Then, because pdoc emits HTML, to get it into the mkdocs build you have to incorporate the HTML as "fake" markdown so that mkdocs will process the markdown files and incorporate it.

It works, but it's a little clunky. Here's the output on my dev machine:

image

There are issue with this. pdoc will use the installed tantivy module. This means that the readthedocs site will have to build the tantivy module which might turn out to be nontrivial. Instead, the pdoc execution could be done as part of CI, commiting the output markdown files (in the `./docs/api` directory). That way, readthedocs only needs to operate on markdown files and no compilation is required.


mocobeta:
Hi @cjrh,
thank you for investigating this in detail!

> There are issue with this. pdoc will use the installed tantivy module. This means that the readthedocs site will have to build the tantivy module which might turn out to be nontrivial. Instead, the pdoc execution could be done as part of CI, commiting the output markdown files (in the ./docs/api directory). That way, readthedocs only needs to operate on markdown files and no compilation is required.

This makes sense to me too.

wallies (approved):

PR #190, opened 2024-01-20
chore: fix module namespace

cjrh:
As identified by @pawamoy, the canonical import path is `tantivy.tantivy`. Our examples work with a single `import tantivy` only because the toplevel module re-imports everything. The correct hierarchy is required for documentation tools to real docstrings correctly.

### DIFF
--- a/src/parser_error.rs
+++ b/src/parser_error.rs
@@ -82,7 +82,7 @@ impl QueryParserErrorIntoPy for tv::query::QueryParserError {
 }
 
 /// Error in the query syntax.
-#[pyclass(frozen)]
+#[pyclass(frozen, module = "tantivy.tantivy")]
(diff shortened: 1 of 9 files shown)


src/parser_error.rs
```diff
@@ -82,7 +82,7 @@ impl QueryParserErrorIntoPy for tv::query::QueryParserError {
 }
 
 /// Error in the query syntax.
-#[pyclass(frozen)]
+#[pyclass(frozen, module = "tantivy.tantivy")]
```
cjrh, line 85: I am not sure whether these changes are required, since these classes are added with `m.add_class` in https://github.com/quickwit-oss/tantivy-py/blob/master/src/lib.rs#L122.

wallies (approved):

Issue #25, opened 2020-10-01
Supporting tokenizer register


acc557:
Currently, the tokenizer is hard-coded to default, it would be better to include some configurable tokenizer for Chinese (tantivy-jieba and cang-jie), Japanese (lindera and tantivy-tokenizer-tiny-segmente) and Korean (lindera + lindera-ko-dic-builder)

https://github.com/tantivy-search/tantivy-py/blob/4ecf7119ea2fc5b3660f38d91a37dfb9e71ece7d/src/schemabuilder.rs#L85

acc557:
@fulmicoton

> Also note that tantivy-py does not come with a japanese tokenizer.
Tantivy has a good and maintained tokenizer called Lindera. If you know rust, you may have to compile your own version of tantivy-py.

I am trying to add LinderaTokenizer to https://github.com/tantivy-search/tantivy-py/blob/master/src/schemabuilder.rs#L85, but I couldn't figure out where 
```
 index
         .tokenizers()
         .register("lang_ja", LinderaTokenizer::new("decompose", ""));
```
should go. Do you have any idea?

fulmicoton:
Anywhere as long as it happens before you index your documents.

Also make sure you declared in the schema that you want to use the tokenizer named "lang_ja" for your japanese fields.

fulmicoton:
Note that these tokenizer typically required shipping a dictionary that is several MB large, so it will not be shipped by default.
Ideally that should be in a different python package, and registration of the tokenizer should be done by the user as suggested by @acc557

zhangchunlin:
What's the progress of adding support configurable tokenizer like tantivy-jieba? This is badly needed for non-ascii text indexing.

fulmicoton:
I don't have time to work on this but any help is welcome.

zhangchunlin:
> 

Could you provide some directions/suggestions we can try? I am willing to do something for this.
Thank you~

adamreichold:
I think a useful approach might be to add optional features to this crate which can be enabled when building it from source using Maturin to include additional tokenizers. Not sure how to best integrate this with pip's optional dependency support though...

cjrh:
I will look at this within the next two weeks or so.

cjrh:
For (my own) future reference the upstream tantivy docs for custom tokenizers is [here](https://docs.rs/tantivy/latest/tantivy/tokenizer/index.html#custom-tokenizers).

cjrh:
[...]

PR #200, opened 2024-01-29
Custom tokenizer support

cjrh:
See #25 

This is an early draft for discussion purposes. It's not ready to be used.  The primary change is that we incorporate the Lindera tantivy tokenizer support under a feature flag.  This means that tantivy-py must be built with that feature flag specified. The included tests have an example of this.  It is possible to run the tests both with and without the Lindera support. We need to decide how we want this to be set up.


Cargo.toml
```diff
@@ -22,7 +22,15 @@ futures = "0.3.26"
 pythonize = "0.20.0"
 serde = "1.0"
 serde_json = "1.0.91"
+# Lindera
+lindera-core = { version = "0.27.2", optional = true }
+lindera-dictionary = { version = "0.27.2", optional = true }
+lindera-tantivy = { version = "0.27.1", optional = true, features = ["ipadic"] }
```
cjrh, line 28: Are there different features that will be required to support?

noxfile.py
```diff
@@ -5,4 +5,17 @@
...
+    session.run("pytest", "-m", "not lindera", *session.posargs)
+
+
+@nox.session(python=["3.8", "3.9", "3.10", "3.11", "3.12"])
+def test_lindera(session):
+    session.install("-rrequirements-dev.txt")
+    session.install(
+        "--no-build-isolation",
+        '--config-settings',
+        'build-args="--features=lindera"',
```
cjrh, line 17: It took a surprisingly long time to discover the correct way to spell how to get this value sent down to the build backend!

src/lindera_tokenizer.rs
```diff
@@ -0,0 +1,16 @@
+use lindera_core::mode::Mode;
+use lindera_dictionary::{
+    load_dictionary_from_config, DictionaryConfig, DictionaryKind,
+};
+use lindera_tantivy::tokenizer::LinderaTokenizer;
+
+pub fn create_tokenizer(mode: Mode) -> LinderaTokenizer {
+    let dictionary_config = DictionaryConfig {
+        kind: Some(DictionaryKind::IPADIC),
```
cjrh, line 9: Should this be a user-configurable option?

src/index.rs
```diff
@@ -244,6 +244,18 @@ impl Index {
...
 
+    /// Register the lindera tokenizer
+    ///
+    /// This will only be available if tantivy-py was built with the "lindera"
+    /// feature.
+    #[cfg(feature = "lindera")]
+    fn register_lindera_tokenizer(
+        &self,
+    ) {
+        let tokenizer = crate::lindera_tokenizer::create_tokenizer(lindera_core::mode::Mode::Normal);
```
cjrh, line 255: Should the mode be a user-configurable option?

Cargo.toml
adamreichold: Since we probably do not want to expose all of Lindera's features as features of tantivy-py, I think we should go for a maximum set here.

noxfile.py
adamreichold: Yes, even with the unified `pyproject.toml` approach, wiring Python builds is still a mess IMHO. Do you remember where in the Maturin documentation you would have expected this? Maybe an improvement is just a PR away?

cjrh: Maturin wasn't too bad, it is pretty easy to find "feature selection" and that shows up under the `build-wheels` section in the maturin docs. Much more difficult to find was how to propagate the feature parameter through the pip interface, as above.  

I had a quick look, I think this section is probably the best place to mention the `build-args` parameter: https://www.maturin.rs/index.html?highlight=pip%20install#python-packaging-basics

src/index.rs
cjrh: I've gone for option 2.

src/lindera_tokenizer.rs
cjrh: Done

Cargo.toml
cjrh: Done

cjrh:
I've added all the lindera features, and also an interface to specify configurable options on registration.  The code is still rough, for example I need to rename quite a few things. I'd love to also figure out how to spell the function signatures correctly, e.g. for the `register_lindera_tokenizer()` function.

cjrh:
It would be much nicer to provide this as an additional package that can be installed. The lack of a stable ABI prevents this, unless we wrap up the Lindera stuff inside a C wrapper and then call that. I'm in two minds about proceeding with this PR as-is, that's why I haven't moved forward with it. Still thinking about it.

cjrh:
Just thinking out loud, we could consider adding some kind of new `ExternalTokenizer` tokenizer, create an instance that encapsulates another tokenizer with a C interface, and that separate tokenizer is wrapped inside the thing with the C interface. Then the latter can be packaged as a wheel and pip installed.  Lots of handwaving but basically add a soapy C layer around the tokenizer registration and usage.

cjrh:
Must be something in the water, because this got published just yesterday. Many projects dealing with the plugin problem: https://www.arroyo.dev/blog/rust-plugin-systems

cjrh:
Ok so the reason I haven't moved forward with this is because I want to explore different solutions that allow a user to pip install additional tokenizers and register them at runtime. It might be possible to create a C abi interface for tokenizers that will allow this to work.

PR #223, opened 2024-03-19
doc: Add mktestdocs, and tutorial for snippets. Fixes #219

### LINKED ISSUE #219: How to use SnippetGenerator and Snippets?
Would it be possible to provide some examples for reference? I would be very grateful.

cjrh:
Quite a nice feature. The `mktestdocs` package automates running all the code in the mkdocs documentation. This means we can keep our docs having actually working code. I found a couple bugs while setting this up.

I've also added a short tutorial for snippets, using the existing test as an example.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -1,5 +1,44 @@
 # Reference
 
+## Setup
+
+We'll use a test index for the examples that follow.
+
+```python
+import os
+from tantivy import SchemaBuilder, Index, Document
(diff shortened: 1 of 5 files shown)


Sidhant29:
This looks good! So essentially the docs code becomes part of the tests so we can see if future releases change the code somehow right ?

Sidhant29 (approved):

PR #233, opened 2024-04-07
Expose Tantivy's FuzzyTermQuery

mocobeta:
Also see #20

This exposes FuzzyTermQuery with some good defaults.

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -54,42 +54,39 @@ impl Query {
     }
 
     /// Construct a Tantivy's FuzzyTermQuery
+    ///
+    /// # Arguments
+    ///
+    /// * `schema` - Schema of the target index.
+    /// * `field_name` - Field name to be searched.
+    /// * `text` - String representation of the query term.
+    /// * `distance` - (Optional) Edit distance you are going to alow. When not specified, the default is 1.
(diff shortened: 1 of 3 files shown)


src/query.rs
```diff
@@ -52,4 +52,46 @@ impl Query {
             inner: Box::new(inner),
         })
     }
+
+    /// Construct a Tantivy's FuzzyTermQuery
+    #[staticmethod]
+    #[pyo3(signature = (schema, field_name, text, distance = 1, transposition_cost_one = true))]
```
mocobeta, line 58: The default values are my preference, but I have no very strong opinions on them.

src/query.rs
```diff
@@ -52,4 +52,46 @@ impl Query {
...
+        );
+        Ok(Query {
+            inner: Box::new(inner),
+        })
+    }
+
+    /// Construct a Tantivy's FuzzyTermQuery of the term prefix
+    #[staticmethod]
+    #[pyo3(signature = (schema, field_name, text, distance = 1, transposition_cost_one = true))]
+    pub(crate) fn fuzzy_term_query_prefix(
```
adamreichold, line 80: I think a Python API would rather use a `prefix = False` keyword argument instead of a separate method here.

adamreichold (approved):

mocobeta:
Hi, I made a mistake in the doc and corrected it in #236.

PR #234, opened 2024-04-07
Expose Tantivy's PhraseQuery

mocobeta:
Also see #20

This exposes PhraseQuery in tantivy-py.

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -1,5 +1,5 @@
 use crate::{make_term, Schema};
-use pyo3::{exceptions, prelude::*, types::PyAny};
+use pyo3::{exceptions, prelude::*, types::PyAny, types::PyList};
 use tantivy as tv;
 
 /// Tantivy's Query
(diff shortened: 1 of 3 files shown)


src/query.rs
```diff
@@ -52,4 +52,50 @@ impl Query {
...
+    #[staticmethod]
+    #[pyo3(signature = (schema, field_name, words, offsets, slop = 0))]
+    pub(crate) fn phrase_query_offset_slop(
+        schema: &Schema,
+        field_name: &str,
+        words: &PyList,
+        offsets: &PyList,
+        slop: u32,
+    ) -> PyResult<Query> {
+        assert!(
```
adamreichold, line 85: Since we return a result anyway, I think we should make this either a just an error instead of a panic (which is rather hard to catch in Python).

Better yet, I would prefer to have this as a single function using the flexibility of Python's dynamic types, e.g.

```rust
/// Construct a Tantivy's PhraseQuery
#[staticmethod]
#[pyo3(signature = (schema, field_name, words, slop = 0))]
pub(crate) fn phrase_query(
    schema: &Schema,
    field_name: &str,
    words: &PyAny,
    slop: u32,
) -> PyResult<Query> {
    let mut terms_with_offset = Vec::<(usize, tv::Term)>::new();
    let mut terms = Vec::<tv::Term>::new();
    for word in words.iter()? {
        if let Ok((offset, word)) = word.extract() {
            let term = make_term(&schema.inner, field_name, w)?;
            terms_with_offset.push((offset, term));
        } else {
            let term = make_term(&schema.inner, field_name, w)?;
            terms.push(term);
        }
    }
    let inner = if !terms_with_offset.is_empty() && terms.is_empty() {
        tv::query::PhraseQuery::new_with_offset(terms_with_offset)
    } else if terms_with_offset.is_empty() && !terms.is_empty() {
        tv::query::PhraseQuery::new(terms)
    } else {
        return Err(PyValueErr::new_err("Provide either offsets for all terms or for none"));
    }
    Ok(Query {
        inner: Box::new(inner),
    })
}
```

(Have neither compiled nor tested this.)

mocobeta: @adamreichold Thank you for your suggestion.

> Since we return a result anyway, I think we should make this either a just an error instead of a panic (which is rather hard to catch in Python).

I think you are totally right - sorry, I have limited experience with actual Rust projects.

cjrh: @mocobeta It is ok - we are learning together 😄

mocobeta: I refined the method signature so that `words` takes either a list of str or a list of tuple[int, str].

mocobeta:
Hi👋 , can I get reviews for this?

cjrh (approved):

PR #236, opened 2024-04-13
Fix fuzzy_term_query() documentation.

mocobeta:
This is a follow-up of #233.
I noticed my understanding of `prefix` parameter was wrong. Fixed the doc and added unit tests that describe fuzzy match behavior.


adamreichold (approved):
Yeah, the test case did not really match the parameter but I did not think it too relevant to hold up the PR. Thanks for going the extra mile and extended the test cases (they are also some of the more useful forms of documentation IMHO).

PR #242, opened 2024-04-21
feat: upgrade tantivy to 0.22

cjrh:
NOTE: This is intentionally still a rough WIP (Sunday night hack time!). I want to get feedback on the macro changes first before putting in more work.

Two changes in tantivy 0.21 -> 0.22 that made the upgrade a little tricky:
- `tantivy::schema::document::Document` is now a trait
- `tantivy::schema::document::Value` is now a trait

We use these a fair bit, especially `Value`.  However, I did some research and experimentation and it seems to me that we can use `TantivyDocument` (implements the now-trait `Document`) and `OwnedValue` (implements the now-trait `Value`).

The tests are passing. Is this the correct way to go?

If this is the right way to go, I'll proceed to clean up the messy bits in the next couple days and get it merged.

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -231,11 +231,8 @@ fn value_to_py(py: Python, value: &Value) -> PyResult<PyObject> {
         }
         Value::Facet(f) => Facet { inner: f.clone() }.into_py(py),
         Value::Array(arr) => {
-            let inner: Vec<_> = arr
-                .iter()
-                .map(|x| value_to_py(py, x))
-                .collect::<PyResult<_>>()?;
-            inner.to_object(py)
+            // TODO implement me
+            unimplemented!();
         }
         Value::Object(obj) => object_to_py(py, obj)?,
         Value::Bool(b) => b.into_py(py),
@@ -245,7 +242,7 @@ fn value_to_py(py: Python, value: &Value) -> PyResult<PyObject> {
 
 fn value_to_string(value: &Value) -> String {
     match value {
-        Value::Null => "null".to_string(),
+        Value::Null => format!("{:?}", value),
         Value::Str(text) => text.clone(),
         Value::U64(num) => format!("{num}"),
         Value::I64(num) => format!("{num}"),
--- a/src/document.rs
+++ b/src/document.rs
@@ -296,6 +296,8 @@ where
 /// necessary for serialization.
 #[derive(Deserialize, Serialize)]
 enum SerdeValue {
+    /// Null
+    Null,
     /// The str type is used for any text information.
     Str(String),
     /// Pre-tokenized str type,
@@ -329,6 +331,7 @@ enum SerdeValue {
 impl From<SerdeValue> for Value {
     fn from(value: SerdeValue) -> Self {
         match value {
+            SerdeValue::Null => Self::Null,
             SerdeValue::Str(v) => Self::Str(v),
             SerdeValue::PreTokStr(v) => Self::PreTokStr(v),
             SerdeValue::U64(v) => Self::U64(v),
--- a/src/document.rs
+++ b/src/document.rs
@@ -14,15 +14,14 @@ use pyo3::{
 
 use chrono::{offset::TimeZone, NaiveDateTime, Utc};
 
-use tantivy::{self as tv, schema::Value};
+use tantivy::{self as tv, schema::document::OwnedValue as Value};
 
 use crate::{facet::Facet, schema::Schema, to_pyerr};
 use serde::{
     ser::SerializeMap, Deserialize, Deserializer, Serialize, Serializer,
 };
-use serde_json::Value as JsonValue;
 use std::{
-    collections::{BTreeMap, HashMap},
+    collections::BTreeMap,
     fmt,
     net::{IpAddr, Ipv6Addr},
     str::FromStr,
@@ -54,7 +53,7 @@ pub(crate) fn extract_value(any: &PyAny) -> PyResult<Value> {
     }
     if let Ok(dict) = any.downcast::<PyDict>() {
         if let Ok(json) = pythonize::depythonize(dict) {
-            return Ok(Value::JsonObject(json));
+            return Ok(Value::Object(json));
         }
     }
     Err(to_pyerr(format!("Value unsupported {any:?}")))
@@ -119,11 +118,11 @@ pub(crate) fn extract_value_for_type(
         tv::schema::Type::Json => {
             if let Ok(json_str) = any.extract::<&str>() {
                 return serde_json::from_str(json_str)
-                    .map(Value::JsonObject)
+                    .map(Value::Object)
                     .map_err(to_pyerr_for_type("Json", field_name, any));
             }
 
-            Value::JsonObject(
+            Value::Object(
                 any.downcast::<PyDict>()
                     .map(|dict| pythonize::depythonize(dict))
                     .map_err(to_pyerr_for_type("Json", field_name, any))?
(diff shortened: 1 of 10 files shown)


src/document.rs
```diff
@@ -14,15 +14,14 @@ use pyo3::{
 
 use chrono::{offset::TimeZone, NaiveDateTime, Utc};
 
-use tantivy::{self as tv, schema::Value};
+use tantivy::{self as tv, schema::document::OwnedValue as Value};
```
cjrh, line 17: I don't like the alias to `Value` here.  If we decide to proceed with the approach in this PR, I'll change all the references to `OwnedValue` throughout the code.

src/document.rs
```diff
@@ -192,32 +191,17 @@ fn extract_value_single_or_list_for_type(
...
+    let dict = PyDict::new(py);
+    for (k, v) in obj.iter() {
+        dict.set_item(k, value_to_py(py, v)?)?;
     }
+    Ok(dict.into())
 }
 
 fn value_to_py(py: Python, value: &Value) -> PyResult<PyObject> {
     Ok(match value {
+        Value::Null => py.None(),
```
cjrh, line 204: `Null` and `Array` have been added into the value types, and `JsonObject` changed to just `Object`.  There are no tests for the new types yet.  I won't add tests for these extra fields in this PR, that can come in a future PR before we do a release with the new version.

src/document.rs
```diff
@@ -243,20 +227,20 @@ fn value_to_py(py: Python, value: &Value) -> PyResult<PyObject> {
             .into_py(py)
         }
         Value::Facet(f) => Facet { inner: f.clone() }.into_py(py),
-        Value::JsonObject(json_object) => {
-            let inner: HashMap<_, _> = json_object
-                .iter()
-                .map(|(k, v)| (k, value_to_object(v, py)))
-                .collect();
+        Value::Array(arr) => {
```
cjrh, line 230: This body is incorrect. I must clean up before merge.

src/document.rs
```diff
@@ -243,20 +227,20 @@ fn value_to_py(py: Python, value: &Value) -> PyResult<PyObject> {
...
+        },
+        Value::Object(obj) => object_to_py(py, obj)?,
         Value::Bool(b) => b.into_py(py),
         Value::IpAddr(i) => (*i).to_string().into_py(py),
     })
 }
 
 fn value_to_string(value: &Value) -> String {
     match value {
+        Value::Null => "null".to_string(),
```
cjrh, line 243: Is there a better string here?

src/document.rs
```diff
@@ -267,10 +251,14 @@ fn value_to_string(value: &Value) -> String {
         Value::PreTokStr(_pretok) => {
             // TODO implement me
             unimplemented!();
-        }
-        Value::JsonObject(json_object) => {
+        },
+        Value::Array(arr) => {
+            let inner: Vec<_> = arr.iter().map(value_to_string).collect();
+            format!("{inner:?}")
```
cjrh, line 257: Instead of this we could also use a JSON representation?

src/index.rs
```diff
@@ -16,7 +16,7 @@ use crate::{
 use tantivy as tv;
 use tantivy::{
     directory::MmapDirectory,
-    schema::{NamedFieldDocument, Term, Value},
+    schema::{NamedFieldDocument, Term, OwnedValue as Value},
```
cjrh, line 19: Reminder: remove alias

src/document.rs
cjrh: (These aliases in this PR are simply to make the changeset small, to get feedback on the approach)

src/index.rs
```diff
@@ -73,7 +73,9 @@ impl IndexWriter {
     /// since the creation of the index.
     pub fn add_document(&mut self, doc: &Document) -> PyResult<u64> {
         let named_doc = NamedFieldDocument(doc.field_values.clone());
-        let doc = self.schema.convert_named_doc(named_doc).map_err(to_pyerr)?;
+        use tv::schema::document::TantivyDocument;
```
cjrh, line 76: move

src/index.rs
```diff
@@ -86,7 +88,10 @@ impl IndexWriter {
     /// The `opstamp` represents the number of documents that have been added
     /// since the creation of the index.
     pub fn add_json(&mut self, json: &str) -> PyResult<u64> {
-        let doc = self.schema.parse_document(json).map_err(to_pyerr)?;
+        use tv::schema::document::TantivyDocument;
```
cjrh, line 91: move import

src/index.rs
```diff
@@ -154,6 +159,11 @@ impl IndexWriter {
         let field = get_field(&self.schema, field_name)?;
         let value = extract_value(field_value)?;
         let term = match value {
+            Value::Null => {
+                return Err(exceptions::PyValueError::new_err(format!(
+                    "Field `{field_name}` is null type not deletable."
+                )))
+            },
```
cjrh, line 166: Not sure about what to do for deletions - is Err correct here (Null, Array, Object)?

src/lib.rs
```diff
@@ -1,5 +1,5 @@
 use ::tantivy as tv;
-use ::tantivy::schema::{Term, Value};
+use ::tantivy::schema::{Term, OwnedValue as Value};
```
cjrh, line 2: Reminder remove alias

src/parser_error.rs
```diff
@@ -319,7 +319,7 @@ impl ExpectedBase64Error {
 
     /// If `true`, the length of the base64 string was invalid.
     fn caused_by_invalid_length(&self) -> bool {
-        matches!(self.decode_error, base64::DecodeError::InvalidLength)
+        matches!(self.decode_error, base64::DecodeError::InvalidLength(_))
```
cjrh, line 322: Check if this is correct (the above was suggested by clippy and it compiles but I haven't checked deeper than that yet)

src/searcher.rs
```diff
@@ -247,9 +248,12 @@ impl Searcher {
     ///
     /// Returns the Document, raises ValueError if the document can't be found.
     fn doc(&self, doc_address: &DocAddress) -> PyResult<Document> {
-        let doc = self.inner.doc(doc_address.into()).map_err(to_pyerr)?;
-        let named_doc = self.inner.schema().to_named_doc(&doc);
-        Ok(Document {
+        // Bring the trait into scope
+        use tantivy::Document;
```
cjrh, line 252: The term `Document` is overloaded unfortunately. This import is needed to bring the trait into scope so that we can call `to_named_doc` on the `TantivyDocument` struct instance.  And then we also return a different `Document` struct, local to tantivy-py.  Need to fix the imports.

Using the same names between tantivy-py and tantivy is generally confusing, and this is a specific example where we have a namespace collision between the trait and a different struct called `Document`.

src/document.rs
Sidhant29: Maybe "None", but I am also wondering if returning an empty string "" is possible, because it may be used in boolean expressions in python ?

src/index.rs
Sidhant29: Should be fine I think,
The best way of dealing with null types in tantivy-py, when I use it, is to define custom null representations for each dataType, like INT_MAX for integer etc.
Whenever encountering null, I swap out the null value for these custom reprs.
I wonder if we should provide that functionality by default, to bridge the 'null' paradigms in python and rust for tantivy.

adamreichold:
> The tests are passing. Is this the correct way to go?

Using `TantivyDocument` and `OnwedValue` is definitely the way to go if everything should continue to work as it does.

Making use of the new flexibility offered by the `Document` and `Value` is usually a follow-up in Rust code bases and I do not think it makes in the Python wrapper because need to handle generic documents without additional knowledge of their structure. It could be that we someday have a special document wrapper that directly uses Python compatible types, i.e. implement `Document` for our local `Document` struct.

src/searcher.rs
adamreichold: Note that you can bring a trait into scope without polluting the namespace, e.g.

```rust
use tantitvy::Document as _;
```

cjrh: Nice TIL 💯

wallies (approved):

src/index.rs
cjrh: I'm not going to change these alias in this PR. If we want to change it, it can come in a future PR.

src/document.rs
cjrh: I'm not going to change these alias in this PR. If we want to change it, it can come in a future PR.

src/lib.rs
cjrh: I'm not going to change these alias in this PR. If we want to change it, it can come in a future PR.

src/document.rs
cjrh: I think you're right. An empty string is the closest thing to null while retaining a string type. It will look the same as a "real" empty string, but I don't think there's much we can here given that the result type of this function is `String`.

src/document.rs
cjrh: I'm going to mark this as unimplemented for now. The current tests all pass.

src/document.rs
cjrh: I decided to rather go with the debug formatter of `OwnedValue` here:

```rust
        Value::Null => format!("{:?}", value),
```

This is because `value_to_string()` is currently used only in the debug formatter for `Document`.

src/document.rs
cjrh: For now this is fine.

src/index.rs
cjrh: For now I'm going to keep it as-is, raising an error. We can rethink this in a future PR if necessary.

src/parser_error.rs
cjrh: I think this is fine. We just match on whatever the length is.

PR #243, opened 2024-04-22
Expose the method of boolean query

alex-au-922:
Added the `Occur` enum, `Query.boolean_query` method and corresponding tests.

Expected usage:

```python
@staticmethod
def boolean_query(subqueries: Sequence[tuple[Occur, Query]]) -> Query:
    ...

Query.boolean_query([(Occur.Must, query1), (Occur.MustNot, query2), (Occur.Should, query3)])
```

<u>**Important Changes**</u>
- Added a wrapper struct `struct OccurQueryPair(Occur, Query)` which lives in the lifetime of the python object `subqueries` for parsing list of tuple of Occur and Query
- Added the `Clone` implementation of `Query` class for the `extract()` method to extract Query class from the PyTuple

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,7 +15,7 @@ mod snippet;
 use document::Document;
 use facet::Facet;
 use index::Index;
-use query::Query;
+use query::{Occur, Query};
 use schema::Schema;
 use schemabuilder::SchemaBuilder;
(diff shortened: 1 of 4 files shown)


cjrh (approved):

cjrh:
Thanks!

PR #244, opened 2024-04-22
Expose Tantivy's DisjunctionMaxQuery

aecio:
Added the `Query.disjunction_max_query` method and corresponding tests.

Expected usage:
```py
query1 = Query.term_query(index.schema, "title", "sea") 
query2 = Query.term_query(index.schema, "title", "mice") 
query = Query.disjunction_max_query([query1, query2], tie_breaker=0.5)

result = index.searcher().search(query, 10)
```

See also issue #20.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,16 +40,21 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: [3.9]
+        python-version: ["3.12"]
+        allow-prereleases: [false]
         include:
           - os: ubuntu-latest
-            python-version: "3.12"
+            python-version: "3.13"
+            allow-prereleases: true
           - os: ubuntu-latest
             python-version: "3.11"
+            allow-prereleases: false
           - os: ubuntu-latest
             python-version: "3.10"
+            allow-prereleases: false
           - os: ubuntu-latest
             python-version: 3.8
+            allow-prereleases: false
     runs-on: "${{ matrix.os }}"
     steps:
       - name: Harden Runner
@@ -72,6 +77,7 @@ jobs:
       - uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d  # 5.1.0
         with:
           python-version: ${{ matrix.python-version }}
+          allow-prereleases: ${{ matrix.allow-prereleases }}
 
       - uses: dtolnay/rust-toolchain@bb45937a053e097f8591208d8e74c90db1873d07
         with:
--- a/noxfile.py
+++ b/noxfile.py
@@ -1,7 +1,7 @@
 import nox
 
 
-@nox.session(python=["3.8", "3.9", "3.10", "3.11", "3.12"])
+@nox.session(python=["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"])
 def test(session):
     session.install("-rrequirements-dev.txt")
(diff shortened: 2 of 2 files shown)


alex-au-922:
Seems our function signatures for both in Rust and in Python (BooleanQuery and DistjunctionMaxQuery) are quite different. Would you mind reviewing my commit as well #243 so that we might come up with a more consistent function signature?

aecio:
It looks like the main difference is the use of PyList vs. Vec for the subqueries parameter. I can try to change it to use Vec instead if you think it's better. I assume PyO3 checks the types automatically in that case, so it's not necessary to check the types manually in the Rust implementation? I'll rebase and try to make these changes tomorrow.

alex-au-922:
I have no preference actually, perhaps @cjrh may provide us some feedback? Whether you think PyList or Vec is better for the project

cjrh:
My preference is to use the Rust types like `Vec` in general, and take advantage of PyO3's automatic conversion, like in the `BooleanQuery` PR.  Thanks @alex-au-922 for having a look at this PR too.

tantivy/tantivy.pyi
```diff
@@ -197,6 +197,10 @@ class Query:
     def all_query() -> Query:
         pass
 
+    @staticmethod
+    def disjunction_max_query(subqueries: list[Query], tie_breaker: Optional[float]) -> Query:
```
mocobeta, line 201:
```suggestion
    def disjunction_max_query(subqueries: list[Query], tie_breaker: Optional[float] = None) -> Query:
```

[nit] According to PyO3 doc, `tie_breaker` would be automatically set to `None` (not tested.) If so, specifying that might be useful for users to save typing a bit.

> As a convenience, functions without a #[pyo3(signature = (...))] option will treat trailing Option<T> arguments as having a default of None.

https://pyo3.rs/v0.21.0/function/signature#trailing-optional-arguments

aecio: It does seem to set `None` automatically. Currently both of the following calls work:
```
query = Query.disjunction_max_query([query1, query2])
query = Query.disjunction_max_query([query1, query2], tie_breaker=0.5)
```
In the latest commit, I added a default value of `None` to the Python signature. Does this address the comment or did I miss the point?

aecio:
Hi all, thanks for the review!
I just pushed a new commit with the following Python signatures and using Rust's `Vec<Query>`:

```py
    @staticmethod
    def boolean_query(subqueries: Sequence[tuple[Occur, Query]]) -> Query:
        pass

    @staticmethod
    def disjunction_max_query(subqueries: Sequence[Query], tie_breaker: Optional[float] = None) -> Query:
        pass
```

```rust
    #[staticmethod]
    pub(crate) fn disjunction_max_query(
        subqueries: Vec<Query>,
        tie_breaker: Option<&PyFloat>,
    ) -> PyResult<Query>
```

Let me know if this addresses the comments.

tantivy/tantivy.pyi
mocobeta: Thank you for including my suggestion in the commit!

cjrh:
I'm fixing the CI issue here: https://github.com/quickwit-oss/tantivy-py/pull/251

When that's merged these tests can rerun.

cjrh (approved):

cjrh:
I approved the PR, just waiting to get a clean test run before merging.

cjrh:
@aecio No, thank _you_ for your contribution 👏🏼 🙇🏼

PR #249, opened 2024-04-23
Expose Tantivy's TermSetQuery

aecio:
This PR adds the `Query.term_set_query` method and corresponding tests.

Expected usage:
```py
terms = ["old", "man"]
query = Query.term_set_query(index.schema, "title", terms)
result = index.searcher().search(query, 10)
```

The new `term_set_query` method has a signature similar to the existing `term_query`:
```py
@staticmethod
def term_query(schema: Schema, field_name: str, field_value: Any, index_option: str = "position") -> Query:
    pass

@staticmethod
def term_set_query(schema: Schema, field_name: str, field_values: Sequence[Any]) -> Query:
    pass
```

See also issue https://github.com/quickwit-oss/tantivy-py/issues/20.

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -89,6 +89,26 @@ impl Query {
         })
     }
 
+    /// Construct a Tantivy's TermSetQuery
+    #[staticmethod]
+    #[pyo3(signature = (schema, field_name, field_values))]
+    pub(crate) fn term_set_query(
+        schema: &Schema,
+        field_name: &str,
+        field_values: Vec<&PyAny>,
+    ) -> PyResult<Query> {
+        let terms = field_values
+            .into_iter()
+            .map(|field_value| {
+                make_term(&schema.inner, field_name, &field_value)
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+        let inner = tv::query::TermSetQuery::new(terms);
+        Ok(Query {
+            inner: Box::new(inner),
+        })
+    }
+
     /// Construct a Tantivy's AllQuery
     #[staticmethod]
     pub(crate) fn all_query() -> PyResult<Query> {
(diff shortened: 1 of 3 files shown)


src/query.rs
```diff
@@ -89,6 +89,32 @@ impl Query {
...
 
+    /// Construct a Tantivy's TermSetQuery
+    #[staticmethod]
+    #[pyo3(signature = (schema, field_name, field_values))]
+    pub(crate) fn term_set_query(
+        schema: &Schema,
+        field_name: &str,
+        field_values: Vec<&PyAny>,
+    ) -> PyResult<Query> {
+        let mut terms: Vec<tv::Term> = Vec::new();
```
cjrh, line 100: Since we know how long `field_values` is, perhaps we can use [with_capacity](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.with_capacity)?  Then we won't need multiple allocations in the loop.  Just an idea.

aecio: Yes, makes total sense. I wasn't aware of this API. I just updated and rebased the PR with the latest master.

adamreichold: Sorry for being late to the party. More often than not, it is preferable to use the built-in iterator adaptors with automatically do things like pre-allocate collections if possible, i.e. in this case collect should achieve the same effect but avoid adding the loop and allocation noise.

adamreichold: Specifically, something like

```rust
let terms = field_values
  .into_iter()
  .map(|field_value| {
    make_term(&schema.inner, field_name, &field_value).map_err(|err| {
      exceptions::PyTypeError::new_err(format!(
          "Cannot create a term from the value: {}",
          field_value
      ))
    })
  })
  .collect::<Result<Vec<_>, _>>()?;
```

should work and be just as efficient.

adamreichold: Much better. Thanks for going the extra mile!

cjrh (approved):

PR #251, opened 2024-04-24
chore: update python versions in CI

cjrh:
Python 3.8 is no longer available on the "macos-latest" target.  This is the current list of supported Python versions: https://devguide.python.org/versions/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,16 +41,20 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         python-version: ["3.12"]
+        allow-prereleases: false
         include:
           - os: ubuntu-latest
             python-version: "3.13"
             allow-prereleases: true
           - os: ubuntu-latest
             python-version: "3.11"
+            allow-prereleases: false
           - os: ubuntu-latest
             python-version: "3.10"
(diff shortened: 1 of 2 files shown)


noxfile.py
```diff
@@ -1,7 +1,7 @@
 import nox
 
 
-@nox.session(python=["3.8", "3.9", "3.10", "3.11", "3.12"])
+@nox.session(python=["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"])
```
wallies, line 4: Shouldn't we remove 3.8 from here

.github/workflows/ci.yml
```diff
@@ -40,10 +40,11 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: [3.9]
+        python-version: ["3.12"]
```
wallies, line 43: Why this change, aren't we removing 3.8?

cjrh: - 3.12 is the current stable version, so that's the one we should test across platforms
- 3.8 is still supported by upstream

cjrh: image

wallies (approved):

Issue #253, opened 2024-04-25
[Doc Update] Tutorial about Building and Executing Queries with Objects


alex-au-922:
> Currently we have a lot of implemented methods regarding the issue https://github.com/quickwit-oss/tantivy-py/issues/20 and I think we should have a proper documentation for parameters accepted and usecases.

yep agree 💯 

> should we add a new section under Building and Executing Queries

> Your suggestions lead me in the direction of:
- we rename the existing `Building and Executing Queries` to `Building and Executing Queries with the Query Parser`
- add a new tutorial called `Building and Executing Queries with the Query Objects`, where we can add tutorial content using the new objects.

What do you think?

We can merge this PR first anyway, and do such docs in a different PR.

_Originally posted by @cjrh in https://github.com/quickwit-oss/tantivy-py/issues/250#issuecomment-2075913283_

alex-au-922:
I think this can definitely introduce users with more advanced usage of the Query Object in case some of the query cannot be parsed with the Query Parser, as mentioned by @cjrh in #250.

But seems a tutorial on the use of Query Objects cannot be very comprehensive (they might fall asleep when reading large chunk of function signatures definitely), will we open another page for the documentation, or just let the user reference the rust version of the equivalent queries?

cjrh:
We will eventually have API documentation in the Reference section of the docs. I'm working on that in another PR.

A really good tutorial would be the example you gave in that other issue, using `BooleanQuery` and the nested disjunction query.  The tutorial doesn't have to be, and probably shouldn't be comprehensive. The main goal is to teach something.

cjrh:
The PR was merged so closing

PR #254, opened 2024-04-25
Docs: Updated the building queries with Query objects section

alex-au-922:
PR for updating the docs with tutorials on using Query objects (issue #253). Not sure if the content of tutorial is enough


cjrh:
It's live: https://tantivy-py.readthedocs.io/en/latest/tutorials/#building-and-executing-queries-with-query-objects

PR #256, opened 2024-04-28
Expose const score query

alex-au-922:
Added ConstScoreQuery class with tests.

Expected usage:
```python
@staticmethod
def const_score_query(query: Query, score: float) -> Query:
    pass

Query.const_score_query(BooleanQuery(...), 0.5)
```


mocobeta (approved):
+1

cjrh (approved):

PR #257, opened 2024-04-28
Expose Tantivy's MoreLikeThisQuery

mocobeta:
#20 

This adds `more_like_this_query()`  to `Query`  class to enable searching similar docs to a target document.
This also includes minor changes to fix/suppress clippy warnings.

### Limitations

In this PR, [`DocumentFields`](https://github.com/quickwit-oss/tantivy/blob/99a59ad37e833e1129f7e1abd25738a0adfae73a/src/query/more_like_this/query.rs#L36) type MLTQuery is not supported. I think it can be a future work.

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -1,4 +1,4 @@
-use crate::{get_field, make_term, to_pyerr, Schema};
+use crate::{get_field, make_term, to_pyerr, DocAddress, Schema};
 use pyo3::{
     exceptions,
     prelude::*,
@@ -100,7 +100,7 @@ impl Query {
         let terms = field_values
             .into_iter()
             .map(|field_value| {
-                make_term(&schema.inner, field_name, &field_value)
+                make_term(&schema.inner, field_name, field_value)
             })
             .collect::<Result<Vec<_>, _>>()?;
         let inner = tv::query::TermSetQuery::new(terms);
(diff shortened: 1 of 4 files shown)


src/query.rs
```diff
@@ -226,4 +226,48 @@ impl Query {
             Err(e) => Err(to_pyerr(e)),
         }
     }
+
+    #[staticmethod]
+    #[pyo3(signature = (doc_address, min_doc_frequency = Some(5), max_doc_frequency = None, min_term_frequency = Some(2), max_query_terms = Some(25), min_word_length = None, max_word_length = None, boost_factor = Some(1.0), stop_words = vec![]))]
+    #[allow(clippy::too_many_arguments)]
```
mocobeta, line 232: I would suppress this warning, but do you have any thoughts?

src/query.rs
```diff
@@ -226,4 +226,48 @@ impl Query {
             Err(e) => Err(to_pyerr(e)),
         }
     }
+
+    #[staticmethod]
+    #[pyo3(signature = (doc_address, min_doc_frequency = Some(5), max_doc_frequency = None, min_term_frequency = Some(2), max_query_terms = Some(25), min_word_length = None, max_word_length = None, boost_factor = Some(1.0), stop_words = vec![]))]
```
mocobeta, line 231: These defaults are copied from the original MoreLikeThis.default()

mocobeta:
Hi👋 , can I get reviews for this?

cjrh:
@mocobeta I'll do it on the weekend. I've had a very busy week at my day job 😄

src/query.rs
cjrh: I'm ok with this.

cjrh (approved):

mocobeta:
Hi @cjrh, thank you for merging!

PR #265, opened 2024-05-04
ci: prep for release

wallies:


.github/workflows/publish.yaml
```diff
@@ -144,7 +147,7 @@ jobs:
 
       - uses: dtolnay/rust-toolchain@bb45937a053e097f8591208d8e74c90db1873d07
         with:
-          toolchain: 1.73.0
+          toolchain: 1.77.2
```
cjrh, line 150: I noticed we also have a `.rust-toolchain.toml` file in the repo. Do you reckon we should bump that as well?

cjrh (approved):

PR #275, opened 2024-05-05
deps: upgrade PyO3 to 0.21

cjrh:
This change upgrades PyO3 to 0.21, which introduces a new API for python object references. The following references have more information:
- Blog post by @davidhewitt : [Replacing PyO3's API without breaking everything downstream](https://polar.sh/davidhewitt/posts/replacing-pyo3-api-pt1)
- [Migration guide for PyO3 0.20 -> 0.21](https://pyo3.rs/v0.21.2/migration#from-020-to-021)

I don't have a super detailed understanding of the upstream changes in PyO3 0.21, but I played whack-a-mole with the compiler and the tests appear to pass. I've pointed out specific areas in the PR where I am unsure.


src/document.rs
```diff
@@ -51,24 +51,22 @@ pub(crate) fn extract_value(any: &PyAny) -> PyResult<Value> {
     if let Ok(b) = any.extract::<Vec<u8>>() {
         return Ok(Value::Bytes(b));
     }
-    if let Ok(dict) = any.downcast::<PyDict>() {
-        if let Ok(json) = pythonize::depythonize(dict) {
-            return Ok(Value::Object(json));
-        }
+    if let Ok(json) = pythonize::depythonize_bound(any.clone()) {
```
cjrh, line 54: The first of several `.clone()` calls in this PR.

If I understand [the documentation](https://pyo3.rs/v0.21.2/migration#migrating-from-the-gil-refs-api-to-boundt) correctly:

> Code will need to add the occasional & to borrow the new smart pointer as &Bound<T> to pass these types around (or use .clone() at the very small cost of increasing the Python reference count)

So based on this my understanding is that `Bound<_, T>.clone()` is cheap (inc refcount) and so this is what I did to get the code to compile. However, I am super unfamiliar with the new PyO3 API so it's entirely possible I've missed something.

I will mark all the new calls to `.clone()` in this PR.

src/document.rs
```diff
@@ -42,7 +42,7 @@ pub(crate) fn extract_value(any: &PyAny) -> PyResult<Value> {
     }
     if let Ok(datetime) = any.extract::<NaiveDateTime>() {
         return Ok(Value::Date(tv::DateTime::from_timestamp_secs(
-            datetime.timestamp(),
+            datetime.and_utc().timestamp(),
```
cjrh, line 45: This is unrelated to the PyO3. Just another warning from chrono that was easy to fix because the warning included the required change to resolve the deprecation warning.

src/document.rs
```diff
@@ -123,16 +121,16 @@ pub(crate) fn extract_value_for_type(
             }
 
             Value::Object(
-                any.downcast::<PyDict>()
-                    .map(|dict| pythonize::depythonize(dict))
-                    .map_err(to_pyerr_for_type("Json", field_name, any))?
+                any.clone().downcast_into::<PyDict>()
+                    .map(|dict| pythonize::depythonize_bound(dict.clone().into_any()))
```
cjrh, line 125: clone

src/document.rs
```diff
@@ -123,16 +121,16 @@ pub(crate) fn extract_value_for_type(
...
+                    .map_err(to_pyerr_for_type("Json", field_name, any.clone()))?
```
cjrh, line 126: (Why are there two calls to `map_err` here? Look at the LHS, they're both the same...?)

src/document.rs
```diff
@@ -123,16 +121,16 @@ pub(crate) fn extract_value_for_type(
...
                     .map_err(to_pyerr_for_type("Json", field_name, any))?,
             )
         }
         tv::schema::Type::IpAddr => {
             let val = any
                 .extract::<&str>()
-                .map_err(to_pyerr_for_type("IpAddr", field_name, any))?;
+                .map_err(to_pyerr_for_type("IpAddr", field_name, any.clone()))?;
```
cjrh, line 133: clone

src/document.rs
```diff
@@ -156,12 +154,12 @@ fn extract_value_single_or_list(any: &PyAny) -> PyResult<Vec<Value>> {
...
 
 fn extract_value_single_or_list_for_type(
-    any: &PyAny,
+    any: Bound<PyAny>,
     field_type: &tv::schema::FieldType,
     field_name: &str,
 ) -> PyResult<Vec<Value>> {
     // Check if a numeric fast field supports multivalues.
-    if let Ok(values) = any.downcast::<PyList>() {
+    if let Ok(values) = any.clone().downcast_into::<PyList>() {
```
cjrh, line 162: clone

src/document.rs
```diff
@@ -170,7 +168,7 @@ fn extract_value_single_or_list_for_type(
                 .unwrap_or(false)
         {
             return Ok(vec![extract_value_for_type(
-                values,
+                values.into_any(),
```
cjrh, line 171: I am not yet sure what these calls like `into_any` do. I added this because it seemed correct and made the compiler happy, but not sure if there's a better way to resolve these situations.

src/document.rs
```diff
@@ -548,7 +546,7 @@ impl Document {
 
     fn extend(
         &mut self,
-        py_dict: &PyDict,
+        py_dict: &Bound<PyDict>,
```
cjrh, line 549: This one takes `&Bound` instead of the many other functions that take `Bound`.  I did this because it makes the caller code more convenient.

src/lib.rs
```diff
@@ -161,10 +161,10 @@ pub(crate) fn get_field(
 pub(crate) fn make_term(
     schema: &tv::schema::Schema,
     field_name: &str,
-    field_value: &PyAny,
+    field_value: Bound<PyAny>,
 ) -> PyResult<tv::Term> {
     let field = get_field(schema, field_name)?;
-    let value = extract_value(field_value)?;
+    let value = extract_value(field_value.clone())?;
```
cjrh, line 167: clone

src/document.rs
adamreichold: > So based on this my understanding is that Bound<_, T>.clone() is cheap (inc refcount) and so this is what I did to get the code to compile. However, I am super unfamiliar with the new PyO3 API so it's entirely possible I've missed something.

Yes, `clone` is indeed cheap as it only increments the reference count. In really performance critical code, I can be helpful to avoid even that cost especially if it ends up ineffective, e.g. temporary clones which increment and then decrement the reference count in a loop. But usually, it's fine.

In this case, the old `depythonize` took a GIL-bound reference `&PyAny`, so it borrowed the reference in Python terms. But this was only possible/safe because we kept a global data structure around which would keep track of any necessary reference count changes and would apply them before releasing the GIL. You can still get the same performance by passing around `&Bound<'_, PyAny>`, but now you can avoid the cost of the global data structure entirely and get full control over any reference count changes.

src/document.rs
```diff
@@ -51,24 +51,22 @@ pub(crate) fn extract_value(any: &PyAny) -> PyResult<Value> {
...
+    any: Bound<PyAny>,
     tv_type: tv::schema::Type,
     field_name: &str,
 ) -> PyResult<Value> {
     // Helper function to create `PyErr`s returned by this function.
     fn to_pyerr_for_type<'a, E: std::error::Error>(
         type_name: &'a str,
         field_name: &'a str,
-        any: &'a PyAny,
+        any: Bound<'a, PyAny>,
```
adamreichold, line 69: This might work using `&'a Bound<'_, PyAny>` as well and thereby avoid some of the short-lived clones? Note that the reference `'py` in the `Bound<'py, T>` is the lifetime for which the GIL is held, not necessarily the lifetime for which a Rust borrow exists. The old GIL-bound references tended to conflate these often distinct lifetimes.

src/document.rs
adamreichold: `into_any` just erases the type information, i.e. it is basically an upcast and the converse to `downcast`. It is necessary because Rust does not have a notion of how `Bound<'_, T>` and `Bound<'_, S>` are related based on how `T` and `S`. In our case, `T` are usually subtypes of `PyAny` and hence the `into_any` upcast is available.

In the past, this was "automatic" because `&'py PyDict` allow getting a short-lived `&'a PyAny` via the `Deref` trait which the compiler calls automatically. We still have [that `Deref`](https://docs.rs/pyo3/latest/pyo3/struct.Bound.html#impl-Deref-for-Bound%3C'py,+T%3E) but it only goes from `Bound<'py, T>` to `&'a Bound<'py, PyAny>` which is insufficient here.

I would say, this is another example where the old PyO3 API conflated the short-lived borrow `'a` and the GIL lifetime `'py` and it somehow worked out via covariance. This was nice, but again the price for this convenience was mandatory global state tracking making everyone's Python-to-Rust calls slower.

src/document.rs
adamreichold: I think one is handling the error from `downcast` and one is handling the error from `depythonize`.

I think the code could be made clearer using either `and_then`

```rust
 any.clone()
    .downcast_into::<PyDict>()
    .map_err(to_pyerr_for_type("Json", field_name, any.clone())
    .and_then(|dict| {
       pythonize::depythonize_bound(dict.into_any())
       .map_err(to_pyerr_for_type("Json", field_name, any.clone())
    })?;
```

or a temporary binding

```rust
let dict = any.clone().downcast_into::<PyDict>().map_err(to_pyerr_for_type("Json", field_name, any.clone()))?;
let obj =  pythonize::depythonize_bound(dict.into_any()).map_err(to_pyerr_for_type("Json", field_name, any))?

Value::Object(obj)
```

but both variants still need two calls to `map_err` as the source `Result` type is different.

(I think the `clone` on the `dict` should be unnecessary as it is not used elsewhere but I have not actually tried it with the borrow checker.)

src/document.rs
adamreichold: If it is sufficient, then `&Bound` should almost always be preferred as it borrows the Python reference instead of bumping the reference count and the double indirection is usually optimized away by the compiler.

cjrh: Ok that's useful info. I wasn't sure because the blog said that the new version of `&PyAny` was `Bound<...>` (without the &).  So I expected clones to be cheap, which they were. I was surprised that there wasn't an explicit recommendation of `&Bound` over `Bound`, but I suppose that's just normal Rust stuff and not worth calling out specifically.  I guess I was just too literal.

image

src/document.rs
adamreichold: Yes, you must clone here.

src/document.rs
cjrh: I mentioned in the other comment, seems to me `depythonize_bound` doesn't take a reference, it wants the owned value.

cjrh: > I think the code could be made clearer

Your `and_then` example looks much better thank you 😄

src/document.rs
cjrh: There are actually two clones here. I know we can't do much about the second one, `dict.clone`, because `depythonize_bound` takes `Bound` and not a reference, but the first one is `any.clone()` and I'm not yet sure if there's a neater way to this than a clone.

src/document.rs
cjrh: Yep, it worked. However I had to not include the inner lifetime parameter at all

```rust
        any: &'a Bound<PyAny>,
```

cjrh:
Huge thanks @adamreichold for your suggestions. I think this is much better now.  I would appreciate another spot check to see if everything looks right?

src/document.rs
```diff
@@ -51,24 +51,22 @@ pub(crate) fn extract_value(any: &PyAny) -> PyResult<Value> {
     if let Ok(b) = any.extract::<Vec<u8>>() {
         return Ok(Value::Bytes(b));
     }
-    if let Ok(dict) = any.downcast::<PyDict>() {
```
adamreichold, line 54: This looks like an unintentional behaviour change, i.e. before only dictionaries were "depythonized" but everything is.

adamreichold (approved):
One question w.r.t. depythonizing non-dicts, otherwise LGTM.

src/document.rs
cjrh: I fixed, I think it's correct?

adamreichold: Agreed.

PR #281, opened 2024-05-11
Expose Range Query (Beta)

alex-au-922:
A large PR on implementing the range query.

**Supported Types**:
- Unsigned Integer
- Integer
- Float
- Date
- IpAddr
(I have no idea how RangeQuery over text works though tantivy supports it, feel free to correct me but now I just make it unavailable)

**Changes**:
- Added `Document` `add_ip_addr` method for feature parity from `from_dict` with `SchemaBuilder`'s schema
- Created a new Enum class of `FieldType` as `Type` might confuse python users with the `typing.Type` class
- Fixed little commenting errors in `Document`'s `add_json` method
- Created the `Query.range_query` method

**Expected Usage**:
```python
@staticmethod
def range_query(
    schema: Schema,
    field_name: str,
    field_type: FieldType,
    lower_bound: _RangeType,
    upper_bound: _RangeType,
    include_lower: bool = True,
    include_upper: bool = True,
) -> Query:
    pass

query = Query.range_query(
    index.schema, 
    "date", 
    FieldType.Date, 
    datetime.datetime(2020, 1, 1), 
    datetime.datetime(2021, 1, 1), 
    include_upper=False
)
```

As the RangeQuery involves a lot of types, I would like to have a more comprehensive review before rolling out to main.


alex-au-922:
I have added back the Makefile for the ease of testing

Makefile
```diff
@@ -0,0 +1,26 @@
+ifeq ($(shell uname),Darwin)
+  EXT := dylib
+else
+  EXT := so
+endif
+
+source_files := $(wildcard src/*.rs)
+
+all: format lint build test
```
adamreichold, line 9: We do have a Noxfile and hence I would recommend to add needed shortcuts to the Noxfile instead of adding a separate means of automation. (I do not really care whether Makefile or Noxfile, just that we use only one of it and not both. For a Python extension, a Noxfile does seem closer to home though...)

alex-au-922: Thanks for your explanation! I have removed the Makefile now

cjrh:
Thanks for this. I'll go through it when I have a chance. Most likely on the weekend.

cjrh (approved):

PR #287, opened 2024-05-18
Add __and__ and __or__ to Query

PSU3D0:
Very new to Tantivy and Rust, so if I missed something apologies!

It would be excellent to combine queries using syntax like `&`, `|`.

I maintain a library, [Docprompt](https://github.com/Page-Leaf/Docprompt), that uses Tantivy internally, and having these operators available would make my usage easier, as well as other developers that are trying to compose queries declarativly using python syntax.

I understand that are a lot of query types in Tantivy. If there are other areas/directions that need to be implemented before we can expose the python API to developers, let me know and I'm happy to implement. Cheers!

PS:

I think a natural followup would be a python-space `QueryParser` that allows Django/SqlAlchemy like ORM usage for constructing queries. To cite the unit tests:

Old:

```
query = index.parse_query("title:men AND body:winter", ["title", "body"])
```

New:

```
query = index.create_query(title="men") & index.create_query(body="winter")
```

This PR would be the very beginnings of a python API in that direction!

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -65,6 +65,26 @@ impl Query {
         Ok(format!("Query({:?})", self.get()))
     }
 
+    pub(crate) fn __and__(&self, other: Query) -> Query {
+        let inner = tv::query::BooleanQuery::from(vec![
+            (tv::query::Occur::Must, self.inner.box_clone()),
+            (tv::query::Occur::Must, other.inner.box_clone()),
+        ]);
+        Query {
+            inner: Box::new(inner),
+        }
+    }
+
+    pub(crate) fn __or__(&self, other: Query) -> Query {
+        let inner = tv::query::BooleanQuery::from(vec![
+            (tv::query::Occur::Should, self.inner.box_clone()),
+            (tv::query::Occur::Should, other.inner.box_clone()),
+        ]);
+        Query {
+            inner: Box::new(inner),
+        }
+    }
+
     /// Construct a Tantivy's TermQuery
     #[staticmethod]
     #[pyo3(signature = (schema, field_name, field_value, index_option = "position"))]
--- a/tests/tantivy_test.py
+++ b/tests/tantivy_test.py
@@ -64,6 +64,32 @@ class TestClass(object):
 
         assert len(result.hits) == 1
 
+    def test_combine_queries(self, ram_index):
+        index = ram_index
+
+        query1 = ram_index.parse_query("title:men", ["title"])
+        query2 = ram_index.parse_query("body:summer", ["body"])
+
+        combined_and = query1 & query2
+
+        searcher = index.searcher()
+        result = searcher.search(combined_and, 10)
+
+        # This is an AND query, so it should return 0 results since summer isn't present
+        assert len(result.hits) == 0
+
+        combined_or = query1 | query2
+
+        result = searcher.search(combined_or, 10)
+
+        assert len(result.hits) == 1
+
+        double_combined = (query1 & query2) | query1
+
+        result = searcher.search(double_combined, 10)
+
+        assert len(result.hits) == 1
+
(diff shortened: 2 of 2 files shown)


src/query.rs
```diff
@@ -65,6 +65,26 @@ impl Query {
         Ok(format!("Query({:?})", self.get()))
     }
 
+    pub(crate) fn __and__(&self, other: Query) -> Query {
+        let inner = tv::query::BooleanQuery::from(vec![
+            (tv::query::Occur::Must, self.inner.box_clone()),
```
adamreichold, line 70: Could we check if `self` or `other` already is a `BooleanQuery` of all-`Must` and apend ourselves to that instead of building trees of `BooleanQuery`?

alex-au-922: Is there any reason why would you want this feature? Complicated syntaxes like `(query1 | query2) & query3` can be very troublesome to implement. Its also common to nest queries in Lucene.

adamreichold: I fear for the reduced efficiency compared to having a flat list of subqueries. AFAIU, Tantivy will not optimize that query tree and its execution will directly mirror the structure of the input query.

alex-au-922: For our current definition,
```rust
pub(crate) struct Query {
    pub(crate) inner: Box<dyn tv::query::Query>,
}
```
the Query is actually accepting a dynamic dispatch `tantivy::query::Query` object. If we need to distinguish the presence of a boolean request, I afraid we need to setup dedicated Query struct but it will lead to big change in the API. The repackaging algorithm on the other hand, is not hard to deal with.

adamreichold: >  If we need to distinguish the presence of a boolean request, I afraid we need to setup dedicated Query struct but it will lead to big change in the API. 

I was rather thinking about downcasting here. I also think we can keep this internal as an optimization instead of changing the API, i.e. downcast to see if we already have a boolean query with matching modifiers and append a clause instead of nesting.

alex-au-922: While I think the idea is fine, the optimization cannot be too deep due to the `Occur` implementation.

Currently we have 3 variants of `Occur`: `Should`, `Must`, `MustNot`. The optimization can be done when we have a tuple of `(Occur, BooleanQuery)` where the boolean query might contain one or more ***groups*** of `Should`, `Must`, `MustNot`.

Their mathematics goes like the symmetric matrix (Take the header as the index 0 `Occur` of the `(Occur, BooleanQuery)` tuple, while the rows `Occur` as the ***groups*** inside the `BooleanQuery`):

||`Should`|`Must`|`MustNot`|
|---|---|---|---|
|`Should`|`Should`|`(Must, [Should, ...])`|`(MustNot, [Should, ...])`|
|`Must`|`(Should, [Must, ...])`|`Must`|`MustNot`|
|`MustNot`|`(Should, [MustNot, ...])`|`MustNot`|`Must`|

This means that we can only ***flatten 1 layer*** those 5 types of `(Occur, InnerOccurGroup)`, am I right?

alex-au-922: > I also think we can keep this internal as an optimization instead of changing the API.

So do you mean you expect whenever `__and__` and `__or__` are used, the above optimization evaluation rule applies while the original `Query.boolean_query` remains the status quo?

adamreichold: > So do you mean you expect whenever __and__ and __or__ are used, the above optimization evaluation rule applies while the original Query.boolean_query remains the status quo?

No, I just want `a & b & c` and `boolean_query([(Must, a), (Must, b), (Must, c)])` to produce the same internal representation using a single boolean query instead of the first form using two boolean queries.

alex-au-922: I think that's what I was going to convey in another way round.

cjrh:
I have merged #531 which makes it easy to combine boolean queries, but is much more explicit. Closing this PR in favour of #531.

PR #288, opened 2024-05-19
feat: Aggregations API

ditsuke:
## What
Adds support for aggregations with a `Searcher.aggregate` method.

## Why
At the moment `tantivy-py` does not support aggregations in any capacity. This PR adds support for all aggregations with an `aggregate` method on `Searcher` which should closely match the existing `search` method. I also started a discussion ([ref]) on the Quickwit Community Discord earlier this month to get feedback on the API. While there was not much feedback, the general consensus I think is that we should also eventually expose `Collector` in `tantivy-py` to allow for more advanced usage and consistency with the Rust API. While I agree with this strongly (part of the reason I started the discussion), I think it is a good idea to start with this simpler API to get aggregation support going like we have with search.

Note that at the moment the aggregations are returned as a dictionary. I'm still considering whether its worth the effort to port all the types over to pyo3 or if there's another better and simpler way. On the other hand, a dict alone might also work just fine. Please feel free to provide feedback on this.

[ref]: https://discord.com/channels/908281611840282624/1146128251144646708/1236608145661624371

### DIFF
--- a/src/searcher.rs
+++ b/src/searcher.rs
@@ -235,13 +235,13 @@ impl Searcher {
         })
     }
 
-    #[pyo3(signature = (_search_query, agg_query))]
+    #[pyo3(signature = (search_query, agg_query))]
     fn aggregate(
         &self,
         py: Python,
-        _search_query: &Query,
+        search_query: &Query,
         agg_query: String,
-    ) -> PyResult<PyObject> {
+    ) -> PyResult<Py<PyDict>> {
         let agg_str = py.allow_threads(move || {
             let agg_collector = AggregationCollector::from_aggs(
                 serde_json::from_str(&agg_query).map_err(to_pyerr)?,
@@ -249,25 +249,31 @@ impl Searcher {
             );
             let agg_res = self
                 .inner
-                .search(
-                    // search_query.get(),
-                    &tv::query::AllQuery,
-                    &agg_collector,
-                )
+                .search(search_query.get(), &agg_collector)
                 .map_err(to_pyerr)?;
 
-            serde_json::to_string(&agg_res).map_err(to_pyerr)
+            println!("agg_res is {:?}", agg_res);
+
+            let ajson = serde_json::to_string(&agg_res).map_err(to_pyerr);
+            println!("ajson is {:?}", ajson);
+            ajson
         })?;
 
-        let locals = [("json_str", agg_str)].into_py_dict(py);
-        let agg_dict_any = py
-            .eval("json.loads(json_str)", None, Some(locals))
-            .map_err(to_pyerr)?;
+        let locals = [("json_str", agg_str)].into_py_dict_bound(py);
+        py.run_bound(
+            r#"
+import json
+agg_dict = json.loads(json_str)
+"#,
+            None,
+            Some(&locals),
+        )
+        .map_err(to_pyerr)?;
 
-        let agg_dict =
-            <PyDict as PyTryFrom>::try_from(agg_dict_any).map_err(to_pyerr)?;
+        let agg_dict_any = locals.get_item("agg_dict")?.unwrap();
+        let agg_dict = agg_dict_any.downcast::<PyDict>()?;
 
-        Ok(agg_dict.into())
+        Ok(agg_dict.clone().unbind())
     }
 
     /// Returns the overall number of documents in the index.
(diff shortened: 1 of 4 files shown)


ditsuke:
Open for review

cjrh:
@ditsuke There are many formatting changes in this PR.  Could you separate the formatting changes into a different PR?  Also, how did you do the formatting, is it just with black, or ruff format or something else?

ditsuke:
@cjrh thanks for taking a look. The formatting changes were made by `ruff`, and while we're on that topic may I suggest deciding on a formatter for the project for consistency? As for now I've reverted those changes (definitely did not belong in this PR). I would like to add the tutorial in a follow up as well.

src/searcher.rs
```diff
@@ -233,6 +235,43 @@ impl Searcher {
         })
     }
 
+    #[pyo3(signature = (search_query, agg_query))]
+    fn aggregate(
+        &self,
+        py: Python,
+        search_query: &Query,
+        agg_query: String,
```
cjrh, line 243: I'm looking at the test, and it looks like the `agg_query` parameter is also a dict. I am wondering whether this parameter should be a PyDict instead.  From the python code side, it would be more natural to pass in a dict here rather than have to first convert it into a string.

cjrh: I see it's a little tricky to convert the PyDict into `Aggregations`. I spent a few minutes exploring some options here. I understand why you went with string input, to make use of the deserialization implementations. Is there a way we can receive a dict here, and then just internally proceed via a serialize+deserialize step?  I think this function really should take a dict parameter instead of a string, but I don't want to hold up this PR waiting for wrapper types for the aggregation machinery.

cjrh: Here's something I came up with that passes the test. It uses Python's `json` module because I couldn't figure out how to convert the `PyDict` into a `HashMap` without getting into the `Aggregation` type support:

```rust
    #[pyo3(signature = (search_query, agg_query))]
    fn aggregate(
        &self,
        py: Python,
        search_query: &Query,
        agg_query: &Bound<PyDict>,
    ) -> PyResult<Py<PyDict>> {
        let json = py.import_bound("json")?;
        let serialized = json.call_method1("dumps", (agg_query,))?;
        let serialized = serialized.extract::<&str>()?;

        let agg_str = py.allow_threads(move || {
            let agg_collector = AggregationCollector::from_aggs(
                serde_json::from_str(&serialized).map_err(to_pyerr)?,
                Default::default(),
            );
            let agg_res = self
                .inner
                .search(search_query.get(), &agg_collector)
                .map_err(to_pyerr)?;

            serde_json::to_string(&agg_res).map_err(to_pyerr)
        })?;

        let locals = [("json_str", agg_str)].into_py_dict_bound(py);
        py.run_bound(
            r#"
import json
agg_dict = json.loads(json_str)
"#,
            None,
            Some(&locals),
        )
        .map_err(to_pyerr)?;

        let agg_dict_any = locals.get_item("agg_dict")?.unwrap();
        let agg_dict = agg_dict_any.downcast::<PyDict>()?;

        Ok(agg_dict.clone().unbind())
    }

```

The idea is that we could move forward with this now, and avoid the work of wrapping the underlying types, while still keeping the query argument a `PyDict`, which is what I think the correct API should be.

cjrh, line 243: A second point is for naming.  Here you have the `search_query` parameter, but in the `search()` function around line 167, the parameter is just called `query`. I think we should keep it the same here for consistency.

Then in the underlying `AggregationCollector::from_aggs()` function, the parameter with the aggregations is called `agg`. I think we should make it the same here, again for consistency against the underlying call.

cjrh: Later we can have a TODO item to use wrapper types for aggregations directly.

ditsuke: Thanks for this suggestion. I very much agree and have had a commit ready to go for a couple days but im blocked out of pushes to my branch for a bit :(. Let me resolve this soon so we can get this in!

ditsuke: On another note, the call_method you used is much neater than my usage of run_bound so I guess we can refactor that part to use the call API as well!

ditsuke: This should be good to go now

ditsuke: Good point! I've renamed these parameters

cjrh (approved):

ditsuke:
Thanks for merging!

PR #349, opened 2024-09-11
Expose version info

wallies:


src/lib.rs
```diff
@@ -90,6 +90,8 @@ fn tantivy(_py: Python, m: &Bound<PyModule>) -> PyResult<()> {
 
     m.add_wrapped(wrap_pymodule!(query_parser_error))?;
 
+    m.add("__version__", tv::version_string())?;
```
cjrh, line 93: I am wondering whether we should expose both our own version number and also the embedded tantivy version number? What do you think?

cjrh (approved):

Issue #371, opened 2024-10-15
Failed to fetch wheel: tantivy==0.22.0 (Python 3.13.0)


mukeshsahnis:
```
os: MacOS 15.0.1 (24A348)
rustc: 1.81.0 (eeb90cda1 2024-09-04)
python: 3.13.0
pip: 24.2
```
```
$ uv pip install -r requirements/requirements-dev.txt
Resolved 157 packages in 1.48s
error: Failed to prepare distributions
  Caused by: Failed to fetch wheel: tantivy==0.22.0
  Caused by: Build backend failed to build wheel through `build_wheel()` with exit status: 1
--- stdout:
Running `maturin pep517 build-wheel -i /Users/foo/Library/Caches/uv/builds-v0/.tmpHCCrBh/bin/python --compatibility off`
--- stderr:
📦 Including license file "/Users/foo/Library/Caches/uv/built-wheels-v3/pypi/tantivy/0.22.0/Zb7PvPJPgL5AdGLS7FeGo/tantivy-0.22.0.tar.gz/LICENSE"
🍹 Building a mixed python/rust project
🔗 Found pyo3 bindings
🐍 Found CPython 3.13 at /Users/foo/Library/Caches/uv/builds-v0/.tmpHCCrBh/bin/python
📡 Using build options bindings from pyproject.toml
💻 Using `MACOSX_DEPLOYMENT_TARGET=11.0` for aarch64-apple-darwin by default
   Compiling pyo3-build-config v0.20.0
   Compiling time v0.3.30
error[E0282]: type annotations needed for `Box<_>`
  --> /Users/foo/.cargo/registry/src/index.crates.io-6f17d22bba15001f/time-0.3.30/src/format_description/parse/mod.rs:83:9
   |
83 |     let items = format_items
   |         ^^^^^
...
86 |     Ok(items.into())
   |              ---- type must be known at this point
   |
   = note: this is an inference error on crate `time` caused by an API change in Rust 1.80.0; update `time` to version `>=0.3.35` by calling `cargo update`

   Compiling pyo3-ffi v0.20.0
   Compiling pyo3 v0.20.0
   Compiling tantivy v0.22.0 (/Users/foo/Library/Caches/uv/built-wheels-v3/pypi/tantivy/0.22.0/Zb7PvPJPgL5AdGLS7FeGo/tantivy-0.22.0.tar.gz)
For more information about this error, try `rustc --explain E0282`.
[...]

PR #401, opened 2024-12-17
Updated `pythonize` and `pyo3`

jamesbraza:
Closes https://github.com/quickwit-oss/tantivy-py/issues/371

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,10 +19,10 @@ chrono = "0.4.40"
 tantivy = "0.24.0"
 itertools = "0.14.0"
 futures = "0.3.31"
-pythonize = "0.21.0"
+pythonize = "0.24.0"
 serde = "1.0"
 serde_json = "1.0.140"
 
 [dependencies.pyo3]
-version = "0.21.0"
+version = "0.24.2"
 features = ["chrono", "extension-module"]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -75,12 +75,6 @@ checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
 name = "bitflags"
-version = "1.3.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
-
-[[package]]
-name = "bitflags"
 version = "2.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "327762f6e5a765692301e5bb513e0d9fef63be86bbc14528052b1cd3e6f03e07"
@@ -450,9 +444,9 @@ dependencies = [
 
 [[package]]
 name = "heck"
-version = "0.4.1"
+version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "95505c38b4572b2d910cecb0281560f54b440a19336cbbcb27bf6ce6adc6f5a8"
+checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
 
 [[package]]
 name = "htmlescape"
@@ -568,16 +562,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cd945864f07fe9f5371a27ad7b52a172b4b499999f1d97574c9fa68373937e12"
 
 [[package]]
-name = "lock_api"
-version = "0.4.11"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3c168f8615b12bc01f9c17e2eb0cc07dcae1940121185446edc3744920e8ef45"
-dependencies = [
- "autocfg",
- "scopeguard",
-]
-
-[[package]]
 name = "log"
 version = "0.4.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -691,29 +675,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "parking_lot"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3742b2c103b9f06bc9fff0a37ff4912935851bee6d36f3c02bcc755bcfec228f"
-dependencies = [
- "lock_api",
- "parking_lot_core",
-]
-
-[[package]]
-name = "parking_lot_core"
-version = "0.9.9"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4c42a9226546d68acdd9c0a280d17ce19bfe27a46bf68784e4066115788d008e"
-dependencies = [
- "cfg-if",
- "libc",
- "redox_syscall",
- "smallvec",
- "windows-targets 0.48.5",
-]
-
-[[package]]
 name = "pin-project-lite"
 version = "0.2.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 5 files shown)


cjrh:
The reason the pickle test fails is because of something that is different in the `pythonize` package in the new version.  When a `Document` object is deserialized, values in the json object that were `i64`, become `u64` after deserialization:

```
Comparing:

{"birth": [Date(2019-08-12T13:00:05Z)], "bytes": [Bytes([97, 98, 99])], "facet": [Facet(Facet(/europe/france))], "float": [F64(1.0)], "integer": [I64(5)], "json": [Object({"a": I64(1), "b": I64(-2)})], "title": [Str("hello world!")], "unsigned": [U64(1)]}

{"birth": [Date(2019-08-12T13:00:05Z)], "bytes": [Bytes([97, 98, 99])], "facet": [Facet(Facet(/europe/france))], "float": [F64(1.0)], "integer": [I64(5)], "json": [Object({"a": U64(1), "b": I64(-2)})], "title": [Str("hello world!")], "unsigned": [U64(1)]}
```

Look carefully at the types in the json object. Before serialization, `I64(1)`, but after deserialization, `U64`.  This seems to only happen with the JSON object. If that is commented (`.add_json`) in the test then the test passes.

jamesbraza:
Yeah I concur with your breakdown that the serialization/deserialization cycle is considering the JSON's `a` key to be unsigned now.

> If that is commented (`.add_json`) in the test then the test passes.

Fwiw, going to `assert orig.to_dict() == pickled.to_dict()` leads to the test passing as well. So it's something within `__eq__` specifically.

---

How are you getting that printed representation with `I64`/`U64`? Using `repr` or `to_dict` I don't see those values.

Also, `Document` has no `__eq__` method, so I am not yet understanding how the `__eq__` check is actually failing.

cjrh:
If the `assert .to_dict()` change makes the test pass, I am ok to proceed with that. On the python side there is no difference between I64 and u64, nor in a json representation of those payloads if they ever get serialized for anything.  What do you reckon?

jamesbraza:
I think it's important for `__eq__` to work across serialization/deserialization.

What do you think of implementing a `Document.__eq__` to invoke `to_dict`:

```python
class Document:
    ...

    def __eq__(self, other) -> bool:
        if not isinstance(other, type(self)):
            return NotImplemented
        return self.to_dict() == other.to_dict()
```

Also, it may be preferred to do this in `document.rs` with the Rust API, as opposed to in Python only.

cjrh:
Overriding `__eq__` makes me nervous.

> I think it's important for __eq__ to work across serialization/deserialization.

Good point. Perhaps we should make a reproducer of the issue using only pythonize, and ask from help from that project.

jamesbraza:
I would be happy to do so, if you can point out how you printed the `Document` such that `U64`/`I64` stuff showed up.

Also, just so I can understand, how do you know the issue is with `pythonize` vs `pyo3`?

cjrh:
> if you can point out how you printed the Document such that U64/I64 stuff showed up.

From memory I put print statements somewhere. Hopefully I still have that code somewhere. I will check tomorrow.

cjrh:
To print out that data, I added a print inside `__richcmp__` like this:

```rust
impl Document {
    fn __richcmp__<'py>(
        &self,
        other: &Self,
        op: CompareOp,
        py: Python<'py>
    ) -> PyResult<Bound<'py, PyAny>> {
        println!("\n\n\nComparing:\n\n{:?}\n\n{:?}", self.field_values, other.field_values);
        match op {
            CompareOp::Eq => {
                let v = (self == other).into_pyobject(py)?.to_owned().into_any();
                Ok(v)
            },
            CompareOp::Ne => {
                let v = (self != other).into_pyobject(py)?.to_owned().into_any();
                Ok(v)
            },
            _ => {
                let v = PyNotImplemented::get(py).to_owned().into_any();
                Ok(v)
            }
        }
    }
}
```

Tracing backwards from here, I also added some prints in the `_internal_from_pythonized()` method:

```rust
impl Document {
    #[staticmethod]
    fn _internal_from_pythonized(serialized: &Bound<PyAny>) -> PyResult<Self> {
        println!("\n\n\nDeserializing: {:?}", serialized);
        let out = pythonize::depythonize(serialized).map_err(to_pyerr);
        let out: Document = out.unwrap();
        println!("\n\n\nDeserialized: {:?}", out);
        println!("\n\n\nDeserialized: {:?}", out.field_values);
        Ok(out)
    }
}
```

So when I run the tests I get this stdout:

```
------------------------------------- Captured stdout call -------------------------------------



Deserializing: {'birth': [{'Date': 1565614805000000000}], 'bytes': [{'Bytes': [97, 98, 99]}], 'facet': [{'Facet': '/europe/france'}], 'float': [{'F64': 1.0}], 'integer': [{'I64': 5}], 'json': [{'Object': {'a': 1, 'b': 2}}], 'title': [{'Str': 'hello world!'}], 'unsigned': [{'U64': 1}]}



Deserialized: Document(birth=[2019-08-12],bytes=[[97, 98, 9],facet=[/europe/fr],float=[1],integer=[5],json=[{"a":1,"b"],title=[hello worl],unsigned=[1])



Deserialized: {"birth": [Date(2019-08-12T13:00:05Z)], "bytes": [Bytes([97, 98, 99])], "facet": [Facet(Facet(/europe/france))], "float": [F64(1.0)], "integer": [I64(5)], "json": [Object({"a": U64(1), "b": U64(2)})], "title": [Str("hello world!")], "unsigned": [U64(1)]}
orig: Document(birth=[2019-08-12],bytes=[[97, 98, 9],facet=[/europe/fr],float=[1],integer=[5],json=[{"a":1,"b"],title=[hello worl],unsigned=[1])
pickled: Document(birth=[2019-08-12],bytes=[[97, 98, 9],facet=[/europe/fr],float=[1],integer=[5],json=[{"a":1,"b"],title=[hello worl],unsigned=[1])



Comparing:

{"birth": [Date(2019-08-12T13:00:05Z)], "bytes": [Bytes([97, 98, 99])], "facet": [Facet(Facet(/europe/france))], "float": [F64(1.0)], "integer": [I64(5)], "json": [Object({"a": I64(1), "b": I64(2)})], "title": [Str("hello world!")], "unsigned": [U64(1)]}

{"birth": [Date(2019-08-12T13:00:05Z)], "bytes": [Bytes([97, 98, 99])], "facet": [Facet(Facet(/europe/france))], "float": [F64(1.0)], "integer": [I64(5)], "json": [Object({"a": U64(1), "b": U64(2)})], "title": [Str("hello world!")], "unsigned": [U64(1)]}
```

jamesbraza:
Okay I made https://github.com/jamesbraza/tantivy-py/pull/1, which:

1. PR description states my current understanding of the problem (newer `pythonize` more aggressively making U64)
2. Poses a somewhat hacky solution, basically checking if one can convert back to I64

I am not happy with this, but I am really in over my head here, so I made https://github.com/davidhewitt/pythonize/issues/80 basically as a SOS.

jamesbraza:
Perhaps `extract_value` at https://github.com/quickwit-oss/tantivy-py/blob/0.22.0/src/document.rs#L37-L39 has something to do with it? I don't see a U64 case, more just talking out loud

cjrh:
@jamesbraza It would be great if you could add a very short reproducer to your issue on pythonize, so that David can easily run it and see the change in type during roundtrip.

cjrh:
I pushed a commit with a new strategy: for the JSON type, deserialize all integers to I64. This solves our problem with the pickle test, and without breaking any handling for toplevel I64/U64 integers. This is probably the best we can do here.

I will also try to update this branch on top of current master.

cjrh:
Successfully merged the main branch in.

Cargo.toml
```diff
@@ -19,10 +19,10 @@ chrono = "0.4.40"
...
 itertools = "0.14.0"
 futures = "0.3.31"
-pythonize = "0.21.0"
+pythonize = "0.23.0"
 serde = "1.0"
 serde_json = "1.0.140"
 
 [dependencies.pyo3]
-version = "0.21.0"
+version = "0.23.5"
```
wallies, line 27: Should we use the same version as pyo3 build config above 0.24.2?

wallies (approved):

Cargo.toml
jamesbraza: Just took care of this, and went ahead and `rebase`'d atop `master` too

jamesbraza:
My apologies for the two month hiatus here, good to see you guys pushed this forward.

To catch myself up and concrete my understanding, just giving a tl;dr on this PR's story:

1. With newer `pythonize`, we found I64 objects after deserialization become U64, breaking `Document.__eq__`.
2. This happens is because newer `pythonize::depythonize` prioritizes deserialization to unsigned ints over signed ints, per https://github.com/davidhewitt/pythonize/pull/69.
3. I posed https://github.com/jamesbraza/tantivy-py/pull/1, which involves expanding `Document.__richcmp__` to recursively cast unsigned integers to signed integers.
4. In https://github.com/davidhewitt/pythonize/issues/80, David poses adding a config option for `depythonize` to prefer signed ints over unsigned ints.
5. Caleb just added a solution of, upon deserialization of `Object`, recursively casting numbers to I64 (if possible), otherwise F64.
    - This is more opinionated than https://github.com/jamesbraza/tantivy-py/pull/1 because it eliminates unsigned values in JSON. It's also more "`tantivy-py` native" by directly using `Object` (`SerdeValue` enum's item for JSON), instead of `Document.__richcmp__`.

In practice this change is backwards compatible ❤️ as far as I am concerned. The only incompatibility is round trip ser/de's an old `Document` will convert any unsigned ints to signed ints in JSON, but that doesn't matter. I am down to ship this 🚢 .

---

In other news, it looks like `pythonize::depythonize_bound` was just removed in `pythonize` version 0.24 in https://github.com/davidhewitt/pythonize/pull/83, so I just updated the code for that. This brings `cargo build` to pass.

Issue #297, opened 2024-06-07
`writer.delete_documents` on tokenized fields behaves unintuitively


lvjg:
add three document，doc_id = test-1， test-2， test-3；
use writer.delete_documents(field_name="doc_id", field_value="test-1")
writer.commit()
writer.wait_merging_threads()
index.reload()

test-1 can still be found through search...

cjrh:
If you have time, we would appreciate a working code snippet that is easy to run to reproduce the issue. See http://www.sscce.org/

Fudge:
I only started using Tantivy-py this weekend so I might be doing something wrong, but this doesn't behave as expected from reading the API:
```
import tantivy

schema_builder = tantivy.SchemaBuilder()
schema_builder.add_text_field("doc", stored=True)
schema = schema_builder.build()
index = tantivy.Index(schema)

writer = index.writer()

docs = ["1", "2", "3"]
for d in docs:
    writer.add_document(
        tantivy.Document(
            doc=d
        )
    )

writer.commit()
index.reload()
searcher = index.searcher()

print("Before delete:")
for d in docs:
    query = tantivy.Query.term_query(schema, "doc", d)
    top_docs = searcher.search(query)
    print(top_docs)

print("Deleting 1 and 2")
print(writer.delete_documents("doc", "1"))
print(writer.delete_documents("doc", "2"))
writer.commit()
index.reload()

print("After delete:")
for d in docs:
    query = tantivy.Query.term_query(schema, "doc", d)
    top_docs = searcher.search(query)
    print(top_docs)

```

Output is:
```
Before delete:
SearchResult(hits: [(0.9808292, DocAddress { segment_ord: 0, doc: 0 })], count: 1)
SearchResult(hits: [(0.9808292, DocAddress { segment_ord: 2, doc: 0 })], count: 1)
SearchResult(hits: [(0.9808292, DocAddress { segment_ord: 1, doc: 0 })], count: 1)
Deleting 1 and 2
8
9
After delete:
SearchResult(hits: [(0.9808292, DocAddress { segment_ord: 0, doc: 0 })], count: 1)
SearchResult(hits: [(0.9808292, DocAddress { segment_ord: 2, doc: 0 })], count: 1)
SearchResult(hits: [(0.9808292, DocAddress { segment_ord: 1, doc: 0 })], count: 1)
```

Fudge:
Refreshing the searcher gives the expected behaviour:
```
print("Deleting 1 and 2")
[...]

PR #435, opened 2025-03-02
fix!: delete docs by text field, use query parser.  fix #432 and #297

### LINKED ISSUE #432: delete_documents fails when using a hypen or an understore as the value
I am using a uuid for my `id` field and when I delete by `id` the delete does not happen (same with underscores). I don't think a hypen or an underscore needs to be escaped?

Removing the hypen or underscore and the delete works fine. I can work around this by removing the hyphen before writing

I was thinking that the `-` is being interpreted as a negation but it is underscores as well.

The Python code to recreate is below;

```
import json
from tantivy import Index, SchemaBuilder

schema_builder = SchemaBuilder()
schema_builder.add_text_field("id", fast=True)
schema = schema_builder.build()

index = Index(schema)

writer = index.writer()

id_str = "test-1"

source_doc = {
    "id": id_str,
}

query = index.parse_query(f"id:{id_str}")
writer.add_json(json.dumps(source_doc))

writer.commit()
writer.wait_merging_threads()
index.reload()
result = index.searcher().search(query)
print(f"Insert: {result}")

writer = index.writer()
writer.delete_documents("id", id_str)
writer.commit()
index.reload()

result = index.searcher().search(query)
print(f"Deleted: {result}")
```

And I get the result

```
Insert: SearchResult(hits: [(0.57536423, DocAddress { segment_ord: 0, doc: 0 })], count: 1)
Deleted: SearchResult(hits: [(0.57536423, DocAddress { segment_ord: 0, doc: 0 })], count: 1)
```

Whereas without the hyphen I get as expected;

```
Insert: SearchResult(hits: [(0.28768212, DocAddress { segment_ord: 0, doc: 0 })], count: 1)
Deleted: SearchResult(hits: [], count: 0)
```

cjrh:


cjrh:
So I did the work here to see what we can do for document deletions on text fields. I am still not 100% convinced that deletion via the query parser is a good idea.  

The _correct_ solution for most people is going to be to use a raw tokenizer on that field. So I think our responsibility at a minimum should be to call that out appropriately in the method docstring, and our docs.

Perhaps our default tokenizer should have `raw` instead of `default`. This would make the common case of deleting via a text field "just work". However, it creates a time bomb because a user could later make a change to the "default" tokenizer and then discover that now deletions act weirdly. So that also feels off.

Another option worth considering is to remove the text option from the `delete_documents` method entirely (return not support or something like that).  Then we could just expose a new method called `delete_documents_by_query` which will do what is in this PR. This API should make it a lot more obvious that all query processing rules will apply in the latter method, including tokenization. I like this idea the best, but it will break users so doing so will require a major version bump.

src/index.rs
```diff
@@ -144,33 +144,94 @@ impl IndexWriter {
...
+            Value::Str(text) => {
+                // Deletion against a text field is a bit more complex, because
+                // the text field is tokenized. We're going to fall back on the
+                // query parser here. The `field_value` parameter can be given
+                // as if it were a query string, and all matching documents will
+                // be deleted. It is important to understand that this can delete
+                // more documents than expected, depending on how the tokenizer
+                // for the field is configured.
+                //
+                // A good example is the use of hypens. Consider a schema like
```
cjrh, line 185: hyphens

cjrh:
Ok, I have a better idea of how to move forward. The most important change is to expose `delete_query`. This should make it obvious what happens when called.  Then we need to decide whether to do anything about `delete_term`. At a minimum, we could add documentation, including the docstring, that emphasizes that tokenization will not be applied.

wallies:
@cjrh i think documentation and docstring for delete_term is the best option

cjrh:
Closing in favour of #456

PR #456, opened 2025-03-23
Add delete-by-query and deprecate delete_documents. fix #432 and #297

### LINKED ISSUE #432: delete_documents fails when using a hypen or an understore as the value

cjrh:
Quick recap: we're trying to solve the very confusing problem of `delete_documents(<field>, <value>)` not applying query parsing rules like tokenization to `<value>`. This is a bit of a footgun currently so we need to improve the dev experience here.

Based on my earlier work and thinking in #435, I think this is the best way to go:

1. Add a new method `delete_documents_by_query()`, which exposes the `delete_query` method from tantivy
2. Add a new method name `delete_documents_by_term`, which does the same things as what `delete_documents` used to do.
3. Retain `delete_documents()`, but make it a wrapper that calls `delete_documents_by_term` and put a deprecation notice on the wrapper.  Then we can look at removing the wrapper in a future version.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -145,7 +145,7 @@ impl IndexWriter {
     }
 
     #[deprecated(
-        note = "This method is deprecated and will be removed in the future. Use either delete_documents_by_tokenized_term, or delete_documents_by_query."
+        note = "This method is deprecated and will be removed in the future. Use either delete_documents_by_term, or delete_documents_by_query."
     )]
     fn delete_documents(
         &mut self,
--- a/src/index.rs
+++ b/src/index.rs
@@ -150,7 +150,7 @@ impl IndexWriter {
         field_name: &str,
         field_value: &Bound<PyAny>,
     ) -> PyResult<u64> {
-        self.delete_documents_by_tokenized_term(field_name, field_value)
+        self.delete_documents_by_term(field_name, field_value)
     }
 
     /// Delete all documents containing a given term.
--- a/src/index.rs
+++ b/src/index.rs
@@ -144,7 +144,29 @@ impl IndexWriter {
         Ok(self.inner()?.commit_opstamp())
     }
 
+    #[deprecated(note = "This method is deprecated and will be removed in the future. Use either delete_documents_by_tokenized_term, or delete_documents_by_query.")]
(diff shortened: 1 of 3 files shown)


cjrh:
We'll wait for #472 first.

wallies (approved):

cjrh:
@andreymal I hadn't noticed that but thanks for bringing it up. I've asked a question on the tantivy discord (in channel #tantivy-dev).

cjrh:
On the discord they said it was "fine" for us to use it 👍🏼 

Link: https://discord.com/channels/908281611840282624/915785344396439552/1413681106825515069

PR #479, opened 2025-05-05
Add Python 3.13 to publish

cjrh:


wallies:
@cjrh should we remove the unsafe skip version as well https://github.com/quickwit-oss/tantivy-py/blob/master/.github/workflows/ci.yml#L52

Issue #80, opened 2023-06-21
Consider building abi3 wheels


adamreichold:
To reduce build effort and improve portability, PyO3 supports building in ["abi3" mode](https://pyo3.rs/v0.19.0/building_and_distribution#py_limited_apiabi3) where Python's stable ABI/API is used which makes the resulting binary wheels forward-compatible beginning at a specified Python version, e.g. wheels using the `abi3-py38` feature are compatible with Python 3.8 and later.

This can cost some performance if there is a lot of back and forth between Rust and Python code, but whether that cost is significant for this extension is hard to say without measuring it.

cjrh:
It's a good idea, we should do this.

cjrh:
fyi @wallies - since we're planning on a new release soon anyway maybe now is a good time to do this.

cjrh:
Had a look. Sadly, `PyDateTime` for now is [not supported](https://docs.rs/pyo3/0.19.2/pyo3/types/struct.PyDateTime.html#) in the limited ABI and we [use it](https://github.com/quickwit-oss/tantivy-py/blob/master/src/document.rs#L9) in `document.rs`.

guilload:
@cjrh, @wallies, would you mind joining the #tantivy-py channel on the [Quickwit Discord server](https://discord.quickwit.io/). There have been many questions regarding `tantivy-py` on the server lately, and we thought you'd be interested in those conversations.

cjrh:
Apparently there is some kind of fallback support for datetime described in this PR: https://github.com/PyO3/pyo3/issues/3633. I need to look at it but perhaps there's a path for us.

adamreichold:
Indeed @Tpt contributed a fully abi3-compatible variant of PyO3's `chrono` integration feature which means that it can be used for these builds with only a slight performance degradation using exactly the same Rust-side API.

PR #488, opened 2025-05-13
document.rs: remove unused PyDateAccess and PyTimeAccess imports

collinanderson:
This is needed for `abi3` builds (#80) which won't have `PyDateAccess` and `PyTimeAccess` according to https://github.com/PyO3/pyo3/pull/4970.


collinanderson:
Ok I'm new to rust and I do see now that it doesn't actually build without importing these traits, so I'll close this PR.

Sorry for the noise!

PR #507, opened 2025-08-04
Context block for IndexWriter

### LINKED ISSUE #506: Context block for IndexWriter
The `IndexWriter` Python class seems like a good candidate for a dunder enter and a dunder exit method.
From the documentation

```py
 with index.writer() as writer:
    writer.add_document(tantivy.Document(
        doc_id=1,
        title=["The Old Man and the Sea"],
        body=["""He was an old man who fished alone in a skiff in the Gulf Stream and he had gone eighty-four days now without taking a fish."""],
  ))
```
would be more ergonomic than:

```py
writer = index.writer()
writer.add_document(tantivy.Document(
    doc_id=1,
    title=["The Old Man and the Sea"],
    body=["""He was an old man who fished alone in a skiff in the Gulf Stream and he had gone eighty-four days now without taking a fish."""],
))
writer.commit()
writer.wait_merging_threads()

```

Bjoern-Rapp:
This is the first time I have written anything in Rust.
Guidance taken from https://github.com/PyO3/pyo3/issues/1205#issuecomment-1164096251

closes #506

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,4 +1,3 @@
 maturin
-puccinialin
 pytest>=4.0
 mktestdocs==0.2.1
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,3 +1,4 @@
 maturin
+puccinialin
 pytest>=4.0
 mktestdocs==0.2.1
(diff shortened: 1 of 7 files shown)


requirements-dev.txt
```diff
@@ -1,3 +1,4 @@
 maturin
+puccinialin
```
cjrh, line 2: We don't want this dev dependency added I think.

cjrh (changes requested):
Thanks, this looks good. Just the comment about removing the additional dev dependency.

Would you be willing to also update the docs about using the writer as a context manager?  I think it could go into the tutorial section, highlighted as an additional way to do it perhaps.

PR #511, opened 2025-08-10
Add explain() method to Query for debugging

a3lem:
The other day I was struggling to understand why my search hits were being ranked in a very different order than I expected[^1]. What would have helped was something like Tantivy's [`explain()`](https://docs.rs/tantivy/latest/tantivy/query/trait.Query.html#method.explain) method. For a given query and a search hit's doc address, it describes which query terms matched and how the terms contributed to the relevance score.

### Changes

- Staying close to Tantivy's design, I added a `Query.explain(searcher, doc_address)` method to generate explanations for search results.
- These explanations are returned as `Explanation` objects, which expose Tantivy's explanation data via a `to_json() -> str` method.
- Updated Python type stubs to include the new method signatures
- Added documentation and usage examples to the tutorials
- Added a test case and ran everything with nox

### Usage
```python
# After performing a search
result = searcher.search(query, 10)
score, doc_address = result.hits[0]

# Get explanation for why this document matched
explanation = query.explain(searcher, doc_address)
print(explanation.to_json())

# [Outputs]
# { "value": 2.1582496,
#  "description": "BooleanClause. sum of ...",
#  "details": [
#    {
#      "value": 0.8781843,
#      "description": "TermQuery, product of...",
# ...
```

[^1]: (In the end I learned that 'term set query' does *not* do what I thought it did, with all term matches contributing an unweighted score of '1.0'.)

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -41,6 +41,7 @@ jobs:
             github.com:443
             pypi.org:443
             uploads.github.com:443
+            release-assets.githubusercontent.com:443
 
       - name: Checkout repository
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2


a3lem:
Not sure if there's anything I can do about the error in the CodeQL workflow, but the linter in the CI shouldn't complain anymore at least.

a3lem:
My pleasure :)Thanks for merging!

PR #526, opened 2025-09-01
Release the GIL when building indexes

petehunt:
We're using tantivy-py in an app that creates a lot of Indexes on the fly. During profiling we noticed that it holds the GIL for a few hundred ms each time, making it difficult to use in an asyncio project. This PR releases the GIL for all methods in `IndexWriter` and `Index`.

# Test plan
* `nox` tests pass
* Upstreamed into our app, observed 200ms thread blocks before, observed none after.
* All testcases in our app pass with this change as well

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -75,11 +75,18 @@ impl IndexWriter {
     /// by the client to align commits with its own document queue.
     /// The `opstamp` represents the number of documents that have been added
     /// since the creation of the index.
-    pub fn add_document(&mut self, doc: &Document) -> PyResult<u64> {
-        let named_doc = NamedFieldDocument(doc.field_values.clone());
-        let doc = TantivyDocument::convert_named_doc(&self.schema, named_doc)
-            .map_err(to_pyerr)?;
-        self.inner()?.add_document(doc).map_err(to_pyerr)
+    pub fn add_document(
+        &mut self,
+        py: Python,
+        doc: &Document,
+    ) -> PyResult<u64> {
+        py.allow_threads(move || {
+            let named_doc = NamedFieldDocument(doc.field_values.clone());
+            let doc =
+                TantivyDocument::convert_named_doc(&self.schema, named_doc)
+                    .map_err(to_pyerr)?;
+            self.inner()?.add_document(doc).map_err(to_pyerr)
+        })
     }
 
     /// Helper for the `add_document` method, but passing a json string.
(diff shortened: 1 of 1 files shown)


src/index.rs
```diff
@@ -75,11 +75,18 @@ impl IndexWriter {
...
-        let doc = TantivyDocument::convert_named_doc(&self.schema, named_doc)
-            .map_err(to_pyerr)?;
-        self.inner()?.add_document(doc).map_err(to_pyerr)
+    pub fn add_document(
+        &mut self,
+        py: Python,
+        doc: &Document,
+    ) -> PyResult<u64> {
+        py.allow_threads(move || {
+            let named_doc = NamedFieldDocument(doc.field_values.clone());
```
petehunt, line 84: should we do this outside of the `allow_threads()` block?

cjrh:
Interesting. Could you add a python test that writes to the same tantivy index from 10 threads in a thread pool?  You can use `concurrent.futures.ThreadPoolExecutor` for example.  The assertion can be that in the end the index has all the required documents.

cjrh:
Oh, and thanks for the submission btw 👍🏼

src/index.rs
```diff
@@ -510,19 +566,23 @@ impl Index {
...
         fuzzy_fields: HashMap<String, (bool, u8, bool)>,
-        py: Python,
     ) -> PyResult<(Query, Vec<PyObject>)> {
         let parser = self.prepare_query_parser(
-            default_field_names,
-            field_boosts,
-            fuzzy_fields,
+            default_field_names.clone(),
+            field_boosts.clone(),
+            fuzzy_fields.clone(),
```
cjrh, line 578: I think these clones are not required? I removed them and it builds.

src/index.rs
```diff
@@ -465,20 +514,27 @@ impl Index {
...
-            fuzzy_fields,
-        )?;
-
-        let query = parser.parse_query(query).map_err(to_pyerr)?;
-
-        Ok(Query { inner: query })
+        let query = query.to_string();
+        let default_field_names = default_field_names.clone();
+        let field_boosts = field_boosts.clone();
+        let fuzzy_fields = fuzzy_fields.clone();
```
cjrh, line 526: These clones are also not necessary it seems, I removed them and it builds.

petehunt:
Hey thanks for the review. I’ll make these changes. I added the defensive clones manually because I wasn’t sure about the guarantees we lose when releasing the GIL. I’ll add that test, update to the new methods and remove the clones. Will probably get to it late this evening

petehunt:
Hi @cjrh, I believe I've made all the changes you've requested.

Additionally, I wrote a concurrency test here: https://gist.github.com/petehunt/eeaabbcf6844c47d601eecfb74e86314, however I couldn't get it to pass with the current release and it doesn't pass in my fork either.
> ValueError: Failed to acquire Lockfile: LockBusy. Some("Failed to acquire index lock. If you are using a regular directory, this means there is already an `IndexWriter` working on this `Directory`, in this process or in a different process.

I poked around a bit to try to solve it but I'm afraid I'm a little out of my depth on it. Let me know if you need anything else!

cjrh:
Actually, I suspect this `LockBusy` behaviour is just expected. Presumably one must catch the LockBusy (or ValueError) and retry after a delay or something along those lines. Anyway I'll have a look at it tomorrow. We might merge your PR without the concurrency test and I'll just work on it myself.

petehunt:
I can take a look at that as well but I suspected it was intended behavior. One thing I tried was releasing the GIL and then acquiring a new per-Index lock, but that didn't resolve the issue either since there were still multiple writers open concurrently.

cjrh:
I'm going to go ahead and merge. I'll do further testing separately. Thanks again for the contribution 🚀

petehunt:
Thank you for maintaining the project!

PR #531, opened 2025-09-06
Helper methods for building boolean queries

cjrh:
This adds `.and_must_match()`, `and_must_not_match()`, and `or_should_match()` methods which produce boolean queries that can be chained in a fluent interface. Some care is taken to keep the chain flatter by reusing a BooleanQuery layer (and simply append the subsequent clauses).

Here's a silly example from the tests:

```python
        index = ram_index
        searcher = index.searcher()

        # Queries for testing
        query_mice = Query.term_query(index.schema, "title", "mice")  # Matches "Of Mice and Men"
        query_old = Query.term_query(index.schema, "title", "old")  # Matches "The Old Man and the Sea"
        query_man = Query.term_query(index.schema, "title", "man")  # Matches "The Old Man and the Sea"

        # "The Old Man and the Sea" contains both "old" and "man"
        # (but with many chains)
        combined_must = (
            query_old
            .and_must_match(query_man)
            .and_must_not_match(query_mice)
        )
        result = searcher.search(combined_must, 10)
        assert len(result.hits) == 1
```

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -99,32 +99,71 @@ impl Query {
 
     /// This is an internal helper method for the BooleanQuery
     /// convenience methods (and_must_match, or_should_match, and_must_not_match).
+    /// It builds a new query in which each query in `others` is added as an
+    /// `other_occur` clause alongside `self`.
+    ///
+    /// When `self` is already a BooleanQuery, the new clauses are appended to
+    /// its clause list where doing so preserves matching semantics, so that
+    /// fluent chains stay flat instead of nesting one level per call:
+    ///
+    /// - Must/MustNot clauses can always be appended, provided the existing
+    ///   `minimum_number_should_match` is carried over. (Tantivy recomputes
+    ///   the minimum to 0 when a Must/MustNot clause is present, which would
+    ///   silently turn existing Should clauses from required-disjunction into
+    ///   optional scoring hints.)
+    /// - Should clauses can only be appended when the existing query is a
+    ///   plain disjunction: all clauses Should, with the default minimum of 1.
+    ///   In any other case, e.g. `a.and_must_match(b).or_should_match(c)`,
+    ///   appending would change which documents match.
+    ///
+    /// In all other cases `self` is nested as a single `self_occur` clause of
+    /// a new BooleanQuery.
     fn combine_with(
         &self,
-        other: Query,
+        others: Vec<Query>,
         self_occur: tv::query::Occur,
         other_occur: tv::query::Occur,
     ) -> Query {
+        use tv::query::Occur;
         type BooleanQuery = tv::query::BooleanQuery;
 
-        let inner: Box<dyn tv::query::Query> = if let Some(boolean_query) =
-            self.inner.downcast_ref::<BooleanQuery>()
-        {
-            let mut subqueries = boolean_query
-                .clauses()
-                .iter()
-                .map(|(occur, subquery)| (*occur, subquery.box_clone()))
-                .collect::<Vec<_>>();
-            subqueries.push((other_occur, other.inner.box_clone()));
-            Box::new(BooleanQuery::new(subqueries))
-        } else {
-            Box::new(BooleanQuery::new(vec![
-                (self_occur, self.inner.box_clone()),
-                (other_occur, other.inner.box_clone()),
-            ]))
-        };
+        if others.is_empty() {
+            return self.clone();
+        }
+        let new_clauses = others.into_iter().map(|query| (other_occur, query.inner));
 
-        Query { inner }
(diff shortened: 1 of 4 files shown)


src/query.rs
```diff
@@ -59,6 +59,36 @@ impl Query {
...
+    fn combine_with(
+        &self,
+        other: Query,
+        self_occur: tv::query::Occur,
+        other_occur: tv::query::Occur,
+    ) -> Query {
+        type BooleanQuery = tv::query::BooleanQuery;
+
+        let inner: Box<dyn tv::query::Query> = if let Some(boolean_query) =
+            self.inner.downcast_ref::<BooleanQuery>()
```
cjrh, line 74: Currently this only checks whether self is a boolean query and will then reuse that to append the new clause; however, it is possible that `other` is also a boolean query. So a further optimization can be made to either append to other (if a BQ), or to fully merge self and other if they're both BQ.

cjrh: (some care must be taken when doing the appends, in case `other_occur` is a MUST NOT)

cjrh: OTOH we may want to intentionally NOT incorporate optimizations for `other`, in light of https://github.com/quickwit-oss/tantivy-py/pull/287#discussion_r1606959868.  It may be desired to only pull things into an existing BooleanQuery when it's on the "left hand side".

cjrh:
One concern that I have for this fluent interface is that it might be a bit annoying to use programmatically.  Something like this might be better:

```python
        # "The Old Man and the Sea" contains both "old" and "man"
        # (but with many chains)
        combined_must = (
            query_old
            .and_must_match([query_man, ...])        # a list of "ands"
            .and_must_not_match([query_mice, ...])   # a list of "and nots"
        )
        result = searcher.search(combined_must, 10)
```

Although I suppose I could support both forms and the fluent interface, using arg unpacking:

```python
        # "The Old Man and the Sea" contains both "old" and "man"
        # (but with many chains)
        combined_must = (
            query_old
            .and_must_match(query_man, <another query>, <another_query>, ...)        # a list of "ands"
            .and_must_not_match(query_mice, <another query>, <another query>, ...)   # a list of "and nots"
        )
        result = searcher.search(combined_must, 10)
```

Then the caller can easily unpack a list into the parameters when calling the `and_must_match()` or `and_must_not_match()` methods.  This seems pretty flexible, I reckon I will go this route.

cjrh:
The varargs suggestion has been implemented.

cjrh:
I like how this has turned out. I'm going to merge.

wallies (commented):
## Review — Helper methods for building boolean queries

### Overview
Adds three fluent methods to `Query` (`and_must_match`, `and_must_not_match`, `or_should_match`) that wrap the existing `boolean_query` primitive for the common AND / AND-NOT / OR cases. All three delegate to a private `combine_with` helper that flattens chains into a single `BooleanQuery` where it's semantically safe, and otherwise nests. CI is green across all platforms and Python versions (incl. 3.14t).

### Correctness — the hard part is handled well
The subtle issue is tantivy's `BooleanQuery`: when a `Must`/`MustNot` clause is present, it recomputes `minimum_number_should_match` to 0, which would silently demote existing `Should` clauses from a required disjunction into optional scoring hints. The PR handles this correctly:

- it reads `get_minimum_number_should_match()` off the existing query and carries it over via `with_minimum_required_clauses`, so `(a OR b).and_must_match(c)` stays `(a OR b) AND c` rather than `c` + optionally-scored a/b;
- it only appends `Should` clauses when the LHS is a *pure* disjunction (`minimum == 1` and every clause `Should`), and nests otherwise.

The `mixed_chains` test (`(old OR mice) AND frankenstein` → 0 hits, `(old AND man) OR mice` → exactly the two expected docs) is precisely the test that would catch a regression here, and it's the most valuable test in the PR.

### Minor points (none blocking)
- **O(n²) cloning on long chains.** Each chained call clones the full existing clause list via `box_clone`, so an N-call chain does roughly N²/2 box-clones. Fine for realistic use (and the 11-deep test); worth at most a one-line awareness comment, not worth optimizing now.
- **No direct test of the flattening optimization.** Coverage verifies behavior, but nothing asserts the chain actually stays flat vs. nested. Since flattening is an internal, behavior-equivalent optimization a test isn't strictly required, though a structural assertion would lock in the perf intent against future refactors.
- **Docs.** The tutorial section is accurate, including the left-to-right precedence caveat and the "flattening is purely internal" note. The `explain()` intro edit (re-pointing at `complex_query`) is a correct fix for the inserted section.

### Verdict
Solid, well-reasoned PR. The tricky `minimum_number_should_match` semantics are correctly preserved, tests cover the dangerous mixed-AND/OR cases, and the docs are clear. No correctness concerns — the points above are optional polish.

wallies (approved):

PR #542, opened 2025-09-16
Add support for custom tokenizers of fast fields

flow3d:
**Why is this feature needed?**
Without it - you can't mark a field as both `fast=True` and to have a custom tokenizer.

The function is minimal, and I've adapted the existing integration test to cover it


flow3d:
Hey @cjrh ! thanks for the quick review.
let me know if there's anything else you need from me before this can get into the main branch?

PR #543, opened 2025-09-18
add phrase_prefix_query support

flow3d:
Added PhrasePrefixQuery support. Tried to keep it similar to phrase_query's implementation for a consistent interface


cjrh:
Thanks!  I'll update #20

PR #548, opened 2025-09-21
export IndexWriter type

flow3d:
Why? it allows for better wrapper creation, explicit type annotation, etc..


cjrh:
Can you expand more on the why? All else being equal I'd prefer to keep the surface area only as big as it needs to be. If there is a specific use-case in mind that requires `IndexWriter` to be exposed then it would be good to describe that. 

For type annotations we already have `IndexWriter` in the `tantivy.pyi` stub file.

cjrh:
nvm I saw @flow3d created the PR not you. Anyhow, I'm recreating in #558.

PR #549, opened 2025-09-21
Document-level score by weight field (fixes #51)

### LINKED ISSUE #51: Adding document based scoring
Hi,

Is there a way to score some documents higher than others? Something like `weight` in whoosh?

For example:
In an ecommerce site, we can have a `popularity` field for each items. And when user searches for something, the popular documents appear on top.

This is already implemented in tantivy core repo:
https://docs.rs/tantivy/0.10.3/tantivy/collector/struct.TopDocs.html#example

I understand that bringing the callback from rust to python may make it slower. Therefore the python interface should be pretty simple. E.g. `searcher.search(query, weight_by='popularity')`

cjrh:


tests/test_document_scoring.py
```diff
@@ -0,0 +1,81 @@
...
+    searcher = index.searcher()
+
+    query_text = "body:banana"
+    query = index.parse_query(query_text)
+    results = searcher.search(query, limit=1)
+    assert len(results.hits) == 1
+    print(results)
+    _, doc_address = results.hits[0]
+    d = index.searcher().doc(doc_address)
+    assert d["id"] == [1]
```
peterbe, line 52: Pardon my ignorance, but does that mean the document with the lesser weight comes before the one with higher?

peterbe: Gawd! I see now. This was the query *without* the `weight_by_field` keyword argument.

PR #559, opened 2025-10-18
Test against free-threaded 3.14t

cjrh:


cjrh:
@wallies What support do I need to add for the wheel builds to work?

PR #574, opened 2025-11-20
feat: Add query_grammar bindings

Darkheir:
Add bindings for the [query-grammar](https://docs.rs/tantivy-query-grammar/0.25.0/tantivy_query_grammar/index.html) functions `parse_query` and `parse_query_lenient`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,6 +8,7 @@ mod facet;
 mod index;
 mod parser_error;
 mod query;
+mod query_grammar;
 mod schema;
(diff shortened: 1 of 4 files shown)


cjrh:
(I'm not saying I _don't_ want to add it - I just want more understanding regarding in which situations it would be useful or necessary)

Darkheir:
To make it clear that those methods are different to the one in the index I could move them to a sub-module called `query_grammar` if you feel it could make a difference

PR #646, opened 2026-04-08
perf: eliminate JSON round-trip in aggregate() and cardinality()

stumpylog:
aggregate() previously used Python's json module to serialize the input dict to a string, passed it through serde_json::from_str, ran the aggregation, then called serde_json::to_string on the result and json.loads to get back a Python dict — four serialization steps total.

cardinality() had the same problem in reverse: it constructed a serde_json::Value internally with serde_json::json!{}, then serialized it to a JSON string and called json.loads to get a Python dict, just to pass it to aggregate() which immediately depythonized it back.

Replace with a private aggregate_value() helper (in a plain impl block, not exposed to Python) that accepts a serde_json::Value directly and returns a Python dict via pythonize — zero JSON strings on either path. The pythonize crate (v0.26.0) is already a dependency used elsewhere in the codebase (query_grammar.rs, document.rs, schema.rs).

Public API is unchanged:
  aggregate(query, dict) -> dict      (depythonize input, aggregate_value, pythonize output)
  cardinality(query, field) -> f64    (serde_json::json! spec, aggregate_value, extract f64)

Tests added:
  test_aggregate_terms  — terms aggregation on a text fast field, verifying
    bucket structure, key types, sort order, and exact counts
  cardinality on body   — text fast field through the new cardinality path

AI disclosure: This was located during profiling in paperless-ngx and Claude Code with Sonnet 4.6 suggested the fix.  The Python JSON module's usage showed up when using cProfile against generated test documents.  (I'm also happy to remove it's authorship, if that is a problem)

### DIFF
--- a/src/searcher.rs
+++ b/src/searcher.rs
@@ -4,6 +4,7 @@ use crate::{document::Document, query::Query, to_pyerr};
 use pyo3::types::PyDict;
 use pyo3::IntoPyObjectExt;
 use pyo3::{basic::CompareOp, exceptions::PyValueError, prelude::*};
+use pythonize::{depythonize, pythonize};
 use serde::{Deserialize, Serialize};
 use tantivy as tv;
 use tantivy::aggregation::AggregationCollector;
@@ -170,6 +171,33 @@ impl SearchResult {
             })
             .collect::<PyResult<_>>()?;
         Ok(ret)
+    }
+}
+
+impl Searcher {
+    /// Execute an aggregation from an already-deserialized spec.
+    /// Shared by `aggregate()` and `cardinality()` so neither needs to
+    /// round-trip through JSON or Python when the spec is already a
+    /// `serde_json::Value`.
+    fn aggregate_value(
+        &self,
+        py: Python,
+        query: &Query,
+        aggs: tv::aggregation::agg_req::Aggregations,
+    ) -> PyResult<Py<PyDict>> {
+        let agg_res = py.detach(move || {
+            let agg_collector =
+                AggregationCollector::from_aggs(aggs, Default::default());
+            self.inner
+                .search(query.get(), &agg_collector)
+                .map_err(to_pyerr)
+        })?;
+
+        pythonize(py, &agg_res)
+            .map_err(to_pyerr)?
+            .downcast_into::<PyDict>()
+            .map(|d| d.unbind())
+            .map_err(Into::into)
     }
 }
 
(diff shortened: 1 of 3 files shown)


wallies:
### Code review

Found 1 issue:

1. Misleading doc comment on public `aggregate()` method. The comment `/// Execute an aggregation from an already-deserialized spec.` looks to be copied from the private `aggregate_value()` helper, but `aggregate()` receives a Python `PyDict`, not an already-deserialized `serde_json::Value`. Either remove the comment or rewrite it to describe the public API accurately.

https://github.com/quickwit-oss/tantivy-py/blob/3e41727d565fd2d49706ac2439ecc66ab204f7e1/src/searcher.rs#L365-L372

🤖 Generated with the help of [Claude Code](https://claude.ai/code)

wallies:
@stumpylog Existing issue but related to this PR

The .pyi stub at tantivy/tantivy.pyi:405 declares:

def aggregate(self, search_query: Query, agg_query: dict) -> dict
But the Rust implementation uses #[pyo3(signature = (query, agg))], parameter names query and agg. So searcher.aggregate(search_query=q, agg_query={...}) would fail at runtime.

stumpylog:
Thanks, I've taken care of both issues

src/searcher.rs
```diff
@@ -333,33 +363,23 @@ impl Searcher {
...
-            serde_json::to_string(&agg_res).map_err(to_pyerr)
-        })?;
-
-        let agg_dict = py_json.call_method1("loads", (agg_str,))?;
-        let agg_dict = agg_dict.downcast::<PyDict>()?;
-
-        Ok(agg_dict.clone().unbind())
+        let agg_spec: serde_json::Value =
+            depythonize(agg.bind(py)).map_err(to_pyerr)?;
+        self.aggregate_value(py, query, agg_spec)
```
cjrh: If you modify the `aggregate_value()` to receive the spec as type `Aggregations`, then you don't need the intermediate `serde::Value`. Something like:

```rust
let aggs: tv::aggregation::agg_req::Aggregations = depythonize(agg.bind(py)).map_err(to_pyerr)?;                
self.aggregate_value(py, query, aggs);
```

stumpylog: I've updated with the suggestion

stumpylog:
Ah, cargo check doesn't check formatting.  I'll resolve it following #641.

stumpylog:
This should be all good to go now.  LMK if there's anything else needed.

wallies (approved):

PR #647, opened 2026-04-09
perf: release GIL in Searcher::doc() and doc_freq()

stumpylog:
Both methods previously held the GIL for the duration of their tantivy calls, blocking all other Python threads while potentially reading from the document store or term dictionary (disk I/O on mmap-backed indexes).

doc(): convert DocAddress to tv::DocAddress before py.detach so we hold no Python references inside the GIL-free region, then read the stored document with the GIL released.

doc_freq(): make_term() still runs with the GIL held (it extracts a Python value), but the tantivy doc_freq() call is moved into py.detach.

Same pattern already used by search() and aggregate().

AI disclosure: Claude Code with Sonnet 4.6 suggested this, though in this case, it wasn't something noticed during integration with paperless-ngx.  (I'm also happy to remove it's authorship, if that is a problem).

Thanks for considering all these PRs.  Even a little taste of Rust has been interesting.  I really should spend some time learning it!

### DIFF
--- a/src/searcher.rs
+++ b/src/searcher.rs
@@ -466,13 +466,14 @@ impl Searcher {
     #[pyo3(signature = (field_name, field_value))]
     fn doc_freq(
         &self,
+        py: Python,
         field_name: &str,
         field_value: &Bound<PyAny>,
     ) -> PyResult<u64> {
-        // Wrap the tantivy Searcher `doc_freq` method to return a PyResult.
+        // make_term() needs the GIL (Python type extraction); doc_freq() does not.
         let schema = self.inner.schema();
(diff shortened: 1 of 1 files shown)


cjrh (approved):
Looks good

cjrh:
Commits need to be signed https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

stumpylog:
Not a problem, signed the commit.  This does require a rebase, so I placed it on top of the latest `master` while doing so.  I'll leave the others alone until they're ready, so I don't need to force push a bunch.

cjrh:
Thanks!

PR #652, opened 2026-04-14
Upgrade tantivy to 0.26.0 and drop Python 3.9

wallies:
## Summary

- Bump `tantivy` dependency to 0.26.0 and the package to 0.26.0.
- Adapt `src/searcher.rs` to the 0.26 API: `TopDocs` is no longer a `Collector` directly (use `.order_by_score()`); `order_by_u64_field` returns `Vec<(Option<u64>, DocAddress)>` (none-highest natural ordering) and is strict about the declared field type. The Python-facing hit value for ordered searches becomes `Optional[int]`.
- Preserve `order_by_field` support for `Date` fast fields by dispatching on schema type and using `order_by_fast_field::<DateTime>` with the sort key mapped through `MonotonicallyMappableToU64::to_u64()`. Non-U64/Date fast fields now raise a clear `ValueError`.
- Drop Python 3.9 (EOL October 2025) from `requires-python`, `noxfile.py`, CI matrices (`ci.yml`, `publish.yaml`), and `ci/deploy_mac.sh`.

New 0.26 features — regex in query grammar, filter aggregation, and other new aggregation types — are already exposed through the existing `parse_query` and `AggregationCollector` passthroughs, so no new Python API is required.

## Test plan

- [x] `cargo build` clean on tantivy 0.26.0 (pre-existing `delete_documents` deprecation warning and `extract_value` unused import are unrelated to this PR)
- [x] `pytest tests/ --ignore=tests/test_docs.py` — 119 passed
- [x] `test_order_by_search_date` still passes (Date dispatch works)
- [x] `test_order_by_search_without_fast_field` updated to assert the new, earlier `ValueError`
- [ ] CI matrix across 3.10–3.14 on Linux/macOS/Windows

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/searcher.rs
+++ b/src/searcher.rs
@@ -8,6 +8,7 @@ use serde::{Deserialize, Serialize};
 use tantivy as tv;
 use tantivy::aggregation::AggregationCollector;
 use tantivy::collector::{Count, MultiCollector, TopDocs};
+use tantivy::columnar::MonotonicallyMappableToU64;
 use tantivy::TantivyDocument;
 
 // Bring the trait into scope. This is required for the `to_named_doc` method.
@@ -30,14 +31,15 @@ enum Fruit {
     #[pyo3(transparent)]
     Score(f32),
     #[pyo3(transparent)]
-    Order(u64),
+    Order(Option<u64>),
 }
 
 impl std::fmt::Debug for Fruit {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             Fruit::Score(s) => f.write_str(&format!("{s}")),
-            Fruit::Order(o) => f.write_str(&format!("{o}")),
+            Fruit::Order(Some(o)) => f.write_str(&format!("{o}")),
+            Fruit::Order(None) => f.write_str("None"),
         }
     }
 }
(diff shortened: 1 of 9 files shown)


src/searcher.rs
```diff
@@ -283,30 +285,73 @@ impl Searcher {
...
-                                })
-                                .collect();
-                            (r, result)
+                    let schema = self.inner.schema();
+                    let field = crate::get_field(&schema, order_by)
+                        .map_err(|e| PyValueError::new_err(e.to_string()))?;
+                    let field_type =
+                        schema.get_field_entry(field).field_type().value_type();
+                    match field_type {
+                        tv::schema::Type::U64 => {
```
cjrh, line 294: Are  there already tests that covet this u64 branch? And what about the date branch?

wallies: Yes test_order_by_search and test_order_by_search_date

wallies: Do you want coverage on PRs?

cjrh (approved):

src/searcher.rs
cjrh: It would be nice to bring in. Historically rust coverage tooling has been patchy but I think it's gotten better recently.

wallies: Yeh ive seen it get better as well

PR #654, opened 2026-04-20
feat: add Searcher.terms_with_prefix for fast prefix-term lookup

stumpylog:
## Summary

Adds `Searcher.terms_with_prefix(field_name, prefix, filter_query=None, limit=None) -> list[tuple[str, int]]`, which walks the FST term dictionary for a text field and returns all terms beginning with a given prefix, together with their document frequencies summed across segments.

The method never fetches stored documents - it reads only the term dictionary and (in the filtered path) posting lists - making it significantly cheaper than issuing a prefix query and aggregating results.

## Motivation

The primary use case is **permission-scoped autocomplete**: suggest indexed terms to a user while restricting counts to documents they are allowed to see. The optional `filter_query` parameter runs a collector over the index first, builds a per-segment set of matching `DocId`s, and then intersects those sets against each term's posting list. This lets a single call produce correctly-scoped term frequencies without a separate suggestion index.

Other use cases this unlocks:

- **Autocomplete / search-as-you-type** - return ranked term suggestions without fetching documents
- **Faceted sidebar counts** - enumerate `(term, count)` pairs for a prefix as a user narrows a filter
- **Vocabulary exploration** - empty prefix + no limit gives the full frequency-ranked vocabulary for a field
- **Query expansion** - enumerate indexed variants before building a broad boolean OR query
- **Corpus debugging** - inspect what terms actually landed in a field after tokenization

## API

```python
searcher.terms_with_prefix(
    field_name: str,
    prefix: str,
    filter_query: Query | None = None,
    limit: int | None = None,
) -> list[tuple[str, int]]
```

- `prefix=""` returns all terms in the field.
- Results are sorted by count descending, then alphabetically for ties.
- Counts are summed across segments, so multi-segment indexes are handled transparently.
- Raises `ValueError` if the field does not exist or is not a text field.

## Implementation notes

- **FST range query** - the upper bound for the prefix scan is computed by `next_prefix_bound`, which increments the last non-`0xFF` byte and truncates. When all bytes are `0xFF`, no FST upper bound can be expressed and the loop falls back to a manual `starts_with` check.
- **Filtered path** - a private `PerSegmentDocSetCollector` implements tantivy's `Collector` trait to gather `HashSet<DocId>` per segment ordinal. This avoids allocating a single global set and lets Phase 2 index directly by segment.
- **GIL release** - the entire body after field validation runs under `py.detach`, consistent with `Searcher::doc` and `doc_freq`.
- **Early exit** - when a `filter_query` is provided but matches no documents, the method returns `[]` immediately without walking any segment's term dictionary.

## Changes

| File | Change |
|---|---|
| `src/searcher.rs` | `next_prefix_bound` helper, `PerSegmentDocSetCollector` + impls, `terms_with_prefix` method on `Searcher` |
| `tantivy/tantivy.pyi` | Type stub for `terms_with_prefix` |
| `tests/tantivy_test.py` | `TestTermsWithPrefix`: 10 tests across 3 fixtures covering prefix match, multi-segment count summing, filter query, empty-filter early exit, limit truncation, sort order, error cases |

## AI disclosure

This PR was developed with Claude Code assistance. All code has been reviewed as best I can.  Overall, most of the additions are tests to ensure the code works as expected.  Those look good to me.

### DIFF
--- a/src/searcher.rs
+++ b/src/searcher.rs
@@ -5,16 +5,93 @@ use pyo3::types::PyDict;
 use pyo3::IntoPyObjectExt;
 use pyo3::{basic::CompareOp, exceptions::PyValueError, prelude::*};
 use serde::{Deserialize, Serialize};
+use std::collections::{HashMap, HashSet};
 use tantivy as tv;
 use tantivy::aggregation::AggregationCollector;
-use tantivy::collector::{Count, MultiCollector, TopDocs};
+use tantivy::collector::{
+    Collector, Count, MultiCollector, SegmentCollector, TopDocs,
+};
 use tantivy::columnar::MonotonicallyMappableToU64;
+use tantivy::schema::{IndexRecordOption, Type};
 use tantivy::TantivyDocument;
+use tantivy::{DocId, DocSet, Score, SegmentOrdinal, TERMINATED};
 
 // Bring the trait into scope. This is required for the `to_named_doc` method.
 // However, tantivy-py declares its own `Document` class, so we need to avoid
 // introduce the `Document` trait into the namespace.
 use tantivy::Document as _;
+
+/// Returns the smallest byte string strictly greater than `prefix`.
+///
+/// Increments the last non-0xFF byte in place and truncates. Returns None
+/// if every byte is 0xFF (caller must fall back to a manual prefix check).
+fn next_prefix_bound(prefix: &[u8]) -> Option<Vec<u8>> {
+    let mut bound = prefix.to_vec();
+    for i in (0..bound.len()).rev() {
+        if bound[i] < 0xFF {
+            bound[i] += 1;
+            bound.truncate(i + 1);
+            return Some(bound);
+        }
+    }
+    None
+}
+
+/// Private collector that gathers matching DocIds per segment.
+///
+/// `merge_fruits` places each segment's HashSet at index `segment_ord` so
+/// `terms_with_prefix` can look up visibility by segment index.
+struct PerSegmentDocSetCollector {
+    num_segments: usize,
+}
+
+struct PerSegmentSegmentCollector {
+    segment_ord: u32,
+    docs: HashSet<DocId>,
+}
+
+impl SegmentCollector for PerSegmentSegmentCollector {
+    type Fruit = (u32, HashSet<DocId>);
+
+    fn collect(&mut self, doc: DocId, _score: Score) {
+        self.docs.insert(doc);
+    }
+
+    fn harvest(self) -> Self::Fruit {
+        (self.segment_ord, self.docs)
+    }
+}
+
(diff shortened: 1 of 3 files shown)


cjrh:
I am wondering whether this feature is more applicable to the upstream tantivy crate? And then tantivy-py would just expose that feature with wrapper code. What are your thoughts?

> The primary use case is permission-scoped autocomplete: suggest indexed terms to a user while restricting counts to documents they are allowed to see. 

The first thing that pops in my head, is that you could build a query that includes a constraint on a particular user. Then the search is scoped.  But looking at it more closely, it looks like this feature doesn't want document hits (search) but instead wants to return terms. So that's a distinction. I guess this is why you can't do autocomplete with the `Searcher.search` interface.

Does the upstream tantivy crate have any features supporting this?

cjrh:
I see tantivy has a method to return the `TermDictionary` here: https://docs.rs/tantivy/latest/tantivy/index/struct.InvertedIndexReader.html#method.terms

cjrh:
The `TermDictionary` itself has a search method: https://docs.rs/tantivy/latest/tantivy/termdict/struct.TermDictionary.html#method.search

cjrh:
Ok I see this PR is using `TermDictionary.range` to produce the sequence of terms. That's a wrapping.

cjrh:
Ok I think is fine. I'm going to accept the idea.

src/searcher.rs
```diff
@@ -626,6 +703,144 @@ impl Searcher {
...
+                        break;
+                    }
+                    let Ok(term_str) = std::str::from_utf8(key) else {
+                        continue;
+                    };
+
+                    let count = match &filter_sets {
+                        None => stream.value().doc_freq,
+                        Some(filter_sets) => {
+                            let term_info = stream.value().clone();
```
cjrh, line 811: Is the `.clone()` necessary? I see that a reference is passed in `read_postings_from_terminfo()`.

src/searcher.rs
```diff
@@ -626,6 +703,144 @@ impl Searcher {
...
+    fn terms_with_prefix(
+        &self,
+        py: Python,
+        field_name: &str,
+        prefix: &str,
+        filter_query: Option<&Query>,
+        limit: Option<usize>,
+    ) -> PyResult<Vec<(String, u32)>> {
+        let schema = self.inner.schema();
+        let field = crate::get_field(&schema, field_name)?;
```
cjrh, line 747: The `&schema` can change to `schema`. Clippy mentions this.

cjrh:
```suggestion
        let field = crate::get_field(schema, field_name)?;
```

cjrh (approved):

src/searcher.rs
cjrh: I'll change this myself after merge.

PR #655, opened 2026-04-20
feat: support unbounded range queries via None bounds

### LINKED ISSUE #529: Unbounded range query
It seems the bindings do not currently allow creating an unbounded range query without the parser? If so, consider this issue as a feature/parity request

```
>>> index.parse_query("score:>=100.0")
Query(RangeQuery { bounds: BoundsRange { lower_bound: Included(Term(field=4, type=F64, 100.0)), upper_bound: Unbounded } })
```

```
>>> tantivy.Query.range_query(
        index.schema,
        field_name="score",
        field_type=tantivy.FieldType.Float,
        lower_bound=100.0,
        upper_bound=None,  # ???
        include_lower=True,
        include_upper=False,  # ???
    )
ValueError: Expected F64 type for field score, got None
```

stumpylog:
Allow passing None for lower_bound or upper_bound in Query.range_query() to produce Tantivy's Unbounded variant, enabling queries like score >= 100 without needing the query parser.

Fixes #529

This fix was developed using Claude Code with Sonnet 4.6


wallies:
### Code review

Found 2 issues:

1. `include_lower`/`include_upper` are silently ignored when the corresponding bound is `None`. Passing e.g. `include_lower=False, lower_bound=None` discards the flag with no error or warning. The combination is semantically meaningless (there is no value to include/exclude at `±∞`), but accepting it silently is a minor usability footgun - consider raising `ValueError` for the contradictory combination.

https://github.com/quickwit-oss/tantivy-py/blob/13b7eeb70276913a009a7c7c8d8623829e3285ad/src/query.rs#L576-L593

2. The both-`None` guard is load-bearing: without it, `tv::query::RangeQuery::weight()` calls `self.field()` → `get_term()` → `.expect("At least one bound must be set")` and panics (tantivy 0.26.0 `range_query.rs:99`; `InvertedIndexRangeQuery::weight()` has the same `.expect` at `range_query.rs:154`). An inline comment would help a future maintainer understand why this guard can't be weakened.

https://github.com/quickwit-oss/tantivy-py/blob/13b7eeb70276913a009a7c7c8d8623829e3285ad/src/query.rs#L570-L576

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

cjrh (approved):

PR #658, opened 2026-04-22
Add support for I64, F64, Bool, and Str fast fields ordering

jborsky:
Building on https://github.com/quickwit-oss/tantivy-py/pull/652, which added `Date` fast field ordering, this PR extends support for ordering by `I64`, `F64`, `Bool`, and `Str` fast fields, matching the full set of types supported by the Rust library.

It leverages the fact that all types except for `Str` are `MonotonicallyMappableToU64`. Thus, we only need to add an `Option<String>` variant to the `Fruit` Enum and the rest will be represented by `Option<u64>`.

Closes https://github.com/quickwit-oss/tantivy-py/issues/643.


wallies:
### Code review

Found 3 issues:

1. The new `run_order_by_fast!` macro returns bit-encoded `u64` values to Python for `i64`, `f64`, and `Date` order keys via `f.map(|v| v.to_u64())`. For `i64`, `-10` becomes `9223372036854775798`; for `f64`, `1.5` becomes the bit-mapped `u64`. Ordering is preserved (that is the point of `MonotonicallyMappableToU64`) but the value exposed as `result.hits[i][0]` is not the original typed value. This pattern was pre-existing for `Date`, but the PR extends it to `I64`/`F64` without documenting the encoding at the Python boundary. The new tests only assert on document ordering, never on `hits[i][0]`, so this silent reinterpret is not covered.

https://github.com/quickwit-oss/tantivy-py/blob/7a766466d2b78e13366d4f46b9f40f6354a7328d/src/searcher.rs#L371-L381

2. The `order_by_field` docstring still reads "Note, this only works for unsigned fields." even though the PR expands support to `I64`, `F64`, `Bool`, `Date`, and `Str`. This was already stale after PR #652 added `Date`, but this PR is the right place to fix it since it actively changes the supported set. 

https://github.com/quickwit-oss/tantivy-py/blob/7a766466d2b78e13366d4f46b9f40f6354a7328d/src/searcher.rs#L218-L223

3. The `str_field` test fixture uses `add_text_field("str_field", fast=True)` rather than a non-tokenized keyword/string field. It works for the single-token values `"apple"` and `"cherry"` because the default tokenizer produces one term per value, but it does not exercise the `Str` fast-field ordering path for realistic multi-token or cased/analyzer-sensitive strings. Consider adding a non-tokenized string field or a multi-word case to catch tokenizer-dependent surprises.

https://github.com/quickwit-oss/tantivy-py/blob/7a766466d2b78e13366d4f46b9f40f6354a7328d/tests/conftest.py#L45-L56

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

jborsky:
Another option would be to add `Fruit` variants for all field types (`OrderI64`, `OrderF64`, `OrderBool`, `OrderDate`) so Python receives the actual values rather than the u64 representation. 

`OrderI64`, `OrderF64`, and `OrderBool` would map directly to their Python counterparts. For `OrderDate`, we can keep it as u64 for simplicity or use serialization to `PyDateTime`.

I followed the `Date` pattern to keep the scope minimal, but happy to implement this if preferred.

wallies:
> Another option would be to add `Fruit` variants for all field types (`OrderI64`, `OrderF64`, `OrderBool`, `OrderDate`) so Python receives the actual values rather than the u64 representation.
> 
> `OrderI64`, `OrderF64`, and `OrderBool` would map directly to their Python counterparts. For `OrderDate`, we can keep it as u64 for simplicity or use serialization to `PyDateTime`.
> 
> I followed the `Date` pattern to keep the scope minimal, but happy to implement this if preferred.

@cjrh what are your thoughts

cjrh:
Will look at on the weekend

tests/conftest.py
```diff
@@ -203,6 +215,36 @@ def create_index_with_ip_addr_field(dir=None):
...
+        doc.add_float("f64_field", 3.14)
+        doc.add_boolean("bool_field", True)
+        doc.add_date("date_field", datetime(2026, 1, 1))
+        doc.add_text("str_field", "cherry")
+        writer.add_document(doc)
+
+        writer.commit()
+        writer.wait_merging_threads()
+        index.reload()
+        return index
```
cjrh: this indentation is too deep

tests/conftest.py
```diff
@@ -328,6 +370,9 @@ def index_with_empty_fast_field():
 def spanish_index():
     return create_spanish_index()
 
+@pytest.fixture(scope="class")
```
cjrh, line 376: Another blank line is needed above this. Module-scope items are separated by two newlines.

cjrh:
> Another option would be to add `Fruit` variants for all field types (`OrderI64`, `OrderF64`, `OrderBool`, `OrderDate`) so Python receives the actual values rather than the u64 representation.
> 
> `OrderI64`, `OrderF64`, and `OrderBool` would map directly to their Python counterparts. For `OrderDate`, we can keep it as u64 for simplicity or use serialization to `PyDateTime`.
> 
> I followed the `Date` pattern to keep the scope minimal, but happy to implement this if preferred.

Yes I think we have to make fruit variants for the specific types. It's going to be too confusing for users otherwise.

For Date, I think we should make a change here. This would be a breaking change.  For the date, I think we should use an i64 representing nanos since epoch.

Here's a handwavy sketch:

```rust
   enum Fruit {                                                                                 
       Score(f32),                                                                              
       OrderU64(Option<u64>),                                                                   
       OrderI64(Option<i64>),                                                                   
       OrderF64(Option<f64>),                                                                   
       OrderBool(Option<bool>),                                                                 
       OrderDate(Option<i64>), // see below                                                     
       OrderStr(Option<String>),                                                                
   }   
```

and then to make the `i64`,

 ```rust                                                                                        
   Fruit::OrderDate(f.map(|dt| dt.into_timestamp_nanos()))                                      
 ```  
 
 Then from Python, users would see:                                                             
                                                                                                
 ```python                                                                                      
   result = searcher.search(query, 10, order_by_field="date_field")                             
                                                                                                
   for date_nanos, doc_address in result.hits:                                                  
       print(date_nanos)                                                                        
 ```                                                                                            
                                                                                                
 Example:                                                                                       
                                                                                                
 ```python                                                                                      
   1704067200000000000  # 2024-01-01T00:00:00Z                                                  
   1767225600000000000  # 2026-01-01T00:00:00Z                                                  
 ```                                                                                            
                                                                                                
 That is much better than the following example u64 values are which an internal encoding and not standard:                                                                      
                                                                                                
 ```python                                                                                      
   10927439236854775808                                                                         
   10990597636854775808                                                                         
 ```                                                                                            
                                                                                                
...because, the Python user can convert it with:                                                             
                                                                                                
 ```python                                                                                      
   from datetime import datetime, timezone                                                      
                                                                                                
   dt = datetime.fromtimestamp(date_nanos / 1_000_000_000, tz=timezone.utc)                     
 ```

jborsky:
Thanks for your comments.

I've done:
- Added the `Fruit` variants.
- Updated the `search()` docstring. 
- Extended the tests to also assert on the actual order key values, not just document order.

cjrh (approved):

PR #681, opened 2026-06-14
test: cover term_query/term_set_query on Date fields with datetime

stumpylog:
Regression tests for passing `datetime` objects to `Query.term_query` and `Query.term_set_query` on `Date` fields.

**Background.** Naive `datetime` queries have worked for a while (the old `Date` branch extracted `NaiveDateTime` at second precision), and #655 added unbounded (`None`) bounds with naive-`datetime` cases in `test_range_query_dates`. Then #666 (`add_date` tzinfo fix) re-routed `extract_value_for_type`'s `Date` branch through the new `pydatetime_to_tv` helper, adding **tz-aware** `datetime` support (the old `NaiveDateTime` extraction rejected any input with `tzinfo`) and sub-second precision. Since `range_query`/`term_query` reach that branch via `make_term[_for_type]`, the query path picked both up — but #666's tests only covered the indexing/round-trip path.

Net: naive range queries are tested (#655); the tz-aware query path and the `term_query`/`term_set_query` date path have no guard.

**Test-only, no behavior change.** Covers, for both queries, naive and tz-aware datetimes (the aware ones exercise the `pydatetime_to_tv` path), plus a non-matching datetime.

Confirmed both new tests fail when `extract_value_for_type`'s `Date` branch is reverted to the pre-#666 `NaiveDateTime` extraction — with the exact `Expected DateTime type for field date, got datetime.datetime(..., tzinfo=...)` error from before that fix. Full `tests/tantivy_test.py` suite passes (147).

AI disclosure (per CONTRIBUTING.md): tests drafted with Claude Code (Opus 4.8); reviewed and verified by me as above.


cjrh (approved):

stumpylog:
I must have missed the message about signing commits somewhere, assuming that was what you meant.  I started with this one and amended with a signed commit

PR #682, opened 2026-06-15
feat: add Index.is_compatible to check index format compatibility

### LINKED ISSUE #337: Implement `is_compatible` to check if an old index can be opened with a new tantivy
https://github.com/quickwit-oss/tantivy/blob/main/src/directory/footer.rs#L104

stumpylog:
Closes #337

Adds `Index.is_compatible(path: str) -> bool` — reports whether an index can be opened by the installed tantivy version, returning `False` instead of raising when the on-disk format version is unsupported.

```python
if Index.is_compatible(path):
    index = Index.open(path)
```

Wraps tantivy's `Footer::is_compatible`, which surfaces as `TantivyError::IncompatibleIndex`: opens cleanly -> `True`, `IncompatibleIndex` -> `False`, other errors (e.g. no index at `path`) -> `ValueError`. Static method taking a path, like `Index.exists`, since an incompatible index can't be opened to call an instance method on.

Includes type stub, tests, and regenerated API docs. `pytest` (167) and `cargo fmt`/`clippy` pass.

Note: the `False` branch isn't unit-tested — it needs an old-format index, and faking one via footer tampering is too brittle. Verified against the tantivy 0.26 source.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cjrh (approved):

stumpylog:
Thanks!  I don't know how often (or if ever yet) there's incompatibility, but I realized this is something we probably needed to check/handle in paperless-ngx, rather than risk an error raised

cjrh:
@stumpylog no worries. See the message about verified signatures though. I'm not sure what you'll have to do to fix that. You might need to look it up.

PR #725, opened 2026-08-18
feat: expose expand_dots_enabled on SchemaBuilder.add_json_field

stumpylog:
## Summary

Split out from #716 per @wallies's review comment there: `JsonObjectOptions::set_expand_dots_enabled()` is public upstream but was never wired into the Python `SchemaBuilder.add_json_field` API. This is a plain binding gap, unrelated to the JSON-subpath term-query work in #716, so it's landing on its own.

- Adds an `expand_dots_enabled` (default `False`) keyword argument to `SchemaBuilder.add_json_field`.
- When set, a `.` in a JSON object key is treated as a path separator — e.g. `{"a.b": "hello"}` becomes queryable as `attrs.a.b` instead of the default escaped `attrs.a\.b` form — matching how `.` already separates path segments in query strings and `Term`s.
- Updates the `tantivy.pyi` stub for `add_json_field` (also fixes a pre-existing gap where `fast` was missing from the stub, noted separately in the #716 review).

## AI disclosure

Per CONTRIBUTING.md's AI policy: the code and test here originated in #716, which was AI-generated (Claude, via Claude Code). I extracted this specific change onto its own branch at @wallies's suggestion; the diff itself is unmodified from #716.

## Test plan

- [x] `cargo fmt --check` passes
- [x] `test_json_field_expand_dots_enabled` covers both the disabled (default, escaped-dot) and enabled (unescaped dotted-path) behavior via `index.parse_query()`

### DIFF
--- a/src/schemabuilder.rs
+++ b/src/schemabuilder.rs
@@ -5,7 +5,7 @@ use pyo3::{exceptions, prelude::*};
 use crate::schema::Schema;
 use std::sync::{Arc, RwLock};
 use tantivy::schema::{
-    self, BytesOptions, DateOptions, IpAddrOptions, INDEXED,
+    self, BytesOptions, DateOptions, IpAddrOptions, JsonObjectOptions, INDEXED,
 };
 
 /// Tantivy has a very strict schema.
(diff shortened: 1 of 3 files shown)


cjrh (approved):

cjrh:
Thanks! 👍🏼

PR #739, opened 2026-09-22
Build musllinux wheels for Linux releases

### LINKED ISSUE #738: Publish musllinux wheels for Alpine
There are no musllinux wheels on PyPI, so installing on Alpine falls back to the sdist and needs a full Rust toolchain. I built 0.26.2 from the sdist in `python:3.13-alpine` (aarch64) and it compiles unpatched and works fine — but it took 3m14s wall / 17m21s CPU, which is what Alpine users pay on every install today.

Would you consider adding musl targets to the release matrix? Happy to open a PR.

hyzyla:
Closes #738


wallies:
@hyzyla can you rebase off main. I just made a change so it makes this testable via PRs, so we know nothing will break.

hyzyla:
> @hyzyla can you rebase off main. I just made a change so it makes this testable via PRs, so we know nothing will break.

done! rebased + added additional test job for musl as well

.github/workflows/publish.yaml
```diff
@@ -387,11 +391,75 @@ jobs:
...
+        with:
+          egress-policy: block
+          allowed-endpoints: >
+            api.github.com:443
+            auth.docker.io:443
+            files.pythonhosted.org:443
+            github.com:443
+            production.cloudflare.docker.com:443
+            pypi.org:443
+            registry-1.docker.io:443
```
wallies: Looks like production.cloudfront.docker.com is needed as well

